PR #17, opened 2026-04-09
fix: auto-refresh OIDC token with proactive + reactive strategy

### LINKED ISSUE #9: feat: auto-refresh OIDC access token during long-lived sessions
## Problem

The CLI authenticates via OIDC at `kontext start` and stores the access token for the session lifetime. The backend `bearerTransport` sends this token on every request (heartbeats, event ingestion, session end).

OIDC access tokens are short-lived (typically 1 hour with Hydra). For coding sessions that run longer than the token TTL, all authenticated requests start failing with `unauthenticated` — heartbeats, event ingestion, and session teardown all silently stop working.

```
2026/04/07 08:43:26 sidecar: heartbeat: unauthenticated: unauthenticated
2026/04/07 08:43:56 sidecar: heartbeat: unauthenticated: unauthenticated
...repeats every 30s for the rest of the session
```

## Desired behavior

The sidecar should transparently refresh the access token before or when it expires, without interrupting the session or requiring user interaction.

### 1. Persist the refresh token

The OIDC auth flow already returns a refresh token alongside the access token. Store it in the session (on disk or keyring) so it's available to the sidecar during the session.

### 2. Proactive refresh in the sidecar

The sidecar knows the token's `expires_at` time. Before expiry (e.g., when 80% of the TTL has elapsed), proactively use the `refresh_token` grant to obtain a new access token. Update the `bearerTransport` atomically so in-flight requests aren't affected.

### 3. Reactive refresh on 401

As a fallback, if any RPC returns `Unauthenticated`, attempt a refresh before retrying. This handles clock skew and edge cases where proactive refresh didn't fire in time.

### 4. Refresh failure → re-auth prompt

If the refresh token itself is expired or revoked, the sidecar can't recover silently. In this case, surface a clear message to the user (e.g., "Session expired, run `kontext login` to re-authenticate") rather than silently failing every 30 seconds.

## Implementation notes

- The `bearerTransport` in `internal/backend/backend.go` currently holds a static token string. It needs to become token-aware: hold a reference to a refreshable token source instead.
- The OIDC client registration must include `refresh_token` in its `grant_types`.
- The refresh token should be stored securely (system keyring or encrypted on disk), not as a plaintext env var.
- Consider making the `Client` struct hold a `TokenSource` interface that the sidecar can swap transparently, similar to `golang.org/x/oauth2.TokenSource`.

tumberger:
## Summary

Fixes #9 — OIDC access tokens expire during long coding sessions, causing heartbeats and event ingestion to silently fail.

### What changed

**Proactive refresh** — The token source checks remaining TTL on every call. When less than 15 minutes remain (~75% of a typical 1h token), it refreshes ahead of time. This prevents latency spikes and avoids concurrent-refresh races on the heartbeat/event paths.

**Reactive refresh (401 retry)** — `bearerTransport.RoundTrip` retries once with a forced token refresh on 401. Handles clock skew, server-side revocation, and edge cases where proactive refresh didn't fire in time.

**Body consumption fix** — The retry path now uses `req.GetBody()` to obtain a fresh request body. Previously, `req.Clone()` shared the original `io.ReadCloser`, so the retry would send an empty body (all ConnectRPC unary RPCs use POST).

**Token persistence** — Refreshed tokens are saved to the system keyring so other processes and subsequent `kontext start` runs see the new token.

### Changes
- `internal/backend/backend.go`: `TokenSource` takes `forceRefresh bool`, `RoundTrip` retries on 401 with fresh body
- `internal/run/run.go`: `newSessionTokenSource` honors `forceRefresh`, `shouldProactiveRefresh` triggers at <15min remaining

## Test plan
- [ ] Run `kontext start --agent claude` for >1h, verify heartbeats continue after token expiry
- [ ] Verify `✓ Token refreshed` appears in stderr when proactive refresh fires
- [ ] Verify 401 retry works by manually expiring the token server-side

---

<a href="https://app.devin.ai/review/kontext-security/kontext-cli/pull/17" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


michiosw (approved):
Looks good now. Retry body handling is fixed, proactive refresh is covered, and checks are green.

PR #48, opened 2026-04-09
feat(cli): check for new releases on startup

tumberger:
## Summary
- Adds a background update checker that queries the GitHub releases API on `kontext start`
- Caches the result in `~/Library/Caches/kontext/version.json` with a 24h TTL
- Prints a one-liner to stderr when a newer version is available, never blocks startup
- Respects `KONTEXT_NO_UPDATE_CHECK` env var for CI / air-gapped environments
- Uses proper semver comparison (not string equality) to avoid false positives
- Sets `User-Agent: kontext-cli/<version>` on GitHub API requests
- No authentication required — uses the public releases endpoint (60 req/hr per IP, 1 req/day with cache)

## Test plan
- [ ] Build with `go build -ldflags "-X main.version=0.0.1" ./cmd/kontext/` and run `kontext start` — should see update notice
- [ ] Run with `KONTEXT_NO_UPDATE_CHECK=1` — no update check
- [ ] Run with `version=dev` — no update check
- [ ] `go test ./internal/update/...` passes
- [ ] `go test -race ./internal/update/...` passes

---

<a href="https://app.devin.ai/review/kontext-dev/kontext-cli/pull/48" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

### DIFF
--- a/cmd/kontext/main.go
+++ b/cmd/kontext/main.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/exec"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -53,7 +54,7 @@ func startCmd() *cobra.Command {
 		Use:   "start [flags] [-- extra-agent-args...]",
 		Short: "Launch an agent with Kontext governance",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			updateDone := update.CheckAsync(version)
+			update.CheckAsync(version)
 			ctx := context.Background()
 			err := run.Start(ctx, run.Options{
 				Agent:        agentName,
@@ -62,7 +63,9 @@ func startCmd() *cobra.Command {
 				ClientID:     auth.DefaultClientID,
 				Args:         args,
 			})
-			<-updateDone
+			if exitErr, ok := err.(*exec.ExitError); ok {
+				os.Exit(exitErr.ExitCode())
+			}
 			return err
 		},
 	}
--- a/cmd/kontext/main.go
+++ b/cmd/kontext/main.go
@@ -50,15 +50,17 @@ func startCmd() *cobra.Command {
 		Use:   "start [flags] [-- extra-agent-args...]",
 		Short: "Launch an agent with Kontext governance",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			update.CheckAsync(version)
+			updateDone := update.CheckAsync(version)
 			ctx := context.Background()
-			return run.Start(ctx, run.Options{
+			err := run.Start(ctx, run.Options{
 				Agent:        agentName,
 				TemplateFile: templateFile,
 				IssuerURL:    auth.DefaultIssuerURL,
 				ClientID:     auth.DefaultClientID,
 				Args:         args,
 			})
+			<-updateDone
+			return err
 		},
 	}
 
--- a/cmd/kontext/main.go
+++ b/cmd/kontext/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/kontext-dev/kontext-cli/internal/hook"
 	"github.com/kontext-dev/kontext-cli/internal/run"
 	"github.com/kontext-dev/kontext-cli/internal/sidecar"
+	"github.com/kontext-dev/kontext-cli/internal/update"
 
 	// Register agent adapters
 	_ "github.com/kontext-dev/kontext-cli/internal/agent/claude"
@@ -49,6 +50,7 @@ func startCmd() *cobra.Command {
 		Use:   "start [flags] [-- extra-agent-args...]",
 		Short: "Launch an agent with Kontext governance",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			update.CheckAsync(version)
 			ctx := context.Background()
 			return run.Start(ctx, run.Options{
 				Agent:        agentName,
--- a/internal/update/update.go
+++ b/internal/update/update.go
@@ -17,6 +17,10 @@ const (
 	cacheTTL = 24 * time.Hour
 )
 
+// githubAPIBase is the base URL for GitHub API requests. Tests override this
+// to point at an httptest server.
+var githubAPIBase = "https://api.github.com"
+
 type cachedVersion struct {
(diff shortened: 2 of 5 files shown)


cmd/kontext/main.go
```diff
@@ -52,14 +53,17 @@ func startCmd() *cobra.Command {
...
 			ctx := context.Background()
-			return run.Start(ctx, run.Options{
+			err := run.Start(ctx, run.Options{
 				Agent:        agentName,
 				TemplateFile: templateFile,
 				IssuerURL:    auth.DefaultIssuerURL,
 				ClientID:     auth.DefaultClientID,
 				Args:         args,
 			})
+			<-updateDone
```
michiosw: Please don't wait on the updater here. Any fast failure or short-lived run now inherits the HTTP timeout from `fetchLatest`, so this is no longer fully off the command path, and `run.Start` still has an `os.Exit` path that skips this wait entirely. Move exit-code propagation out of `run.Start`, then keep the updater opportunistic from `main`.

internal/update/update.go
```diff
@@ -0,0 +1,164 @@
...
+func fetchLatest(currentVersion string) string {
+	client := &http.Client{Timeout: 3 * time.Second}
+	url := fmt.Sprintf("https://api.github.com/repos/%s/releases/latest", repo)
+
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return ""
+	}
+	req.Header.Set("Accept", "application/vnd.github+json")
+	req.Header.Set("User-Agent", "kontext-cli/"+currentVersion)
```
michiosw, line 140: Can we pin `X-GitHub-Api-Version` on this request? GitHub's REST API is versioned in 2026, and unversioned clients stay on the legacy default. We should declare the version we test against so this updater doesn't silently drift later.

internal/update/update_test.go
```diff
@@ -0,0 +1,154 @@
...
+	// Mock GitHub API.
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if ua := r.Header.Get("User-Agent"); ua == "" {
+			t.Error("expected User-Agent header")
+		}
+		json.NewEncoder(w).Encode(map[string]string{"tag_name": "v0.3.0"})
+	}))
+	defer srv.Close()
+
+	// Override the repo URL by testing the internal function directly isn't
```
michiosw: Please refactor this so the test calls `fetchLatest` instead of reimplementing its request/decode path. Right now this still passes if the production function breaks its request wiring, status handling, or URL setup, so it doesn't actually protect the behavior we care about.

michiosw (changes requested):
PR summary:
This adds a GitHub-backed release notifier to `kontext start`, plus a local cache and opt-out env var. The intent is good and the implementation is lean, but there is still one lifecycle bug that breaks the 'background and non-blocking' contract on real exits.

Action items:
- Correctness: move exit-code propagation out of `run.Start` so the update check lifecycle is owned in `main`, or stop waiting on the goroutine entirely. Right now non-zero agent exits bypass the wait, and fast-fail paths still inherit the updater timeout.
- Design/Maintainability: pin `X-GitHub-Api-Version` on the release request. GitHub REST is versioned in 2026, so we should declare the contract we test against instead of floating on the legacy default.
- Tests: refactor the fetch path so `fetchLatest` itself is injectable and unit-testable. The current HTTP test reimplements the request/decode flow instead of exercising the production function.
- Documentation: document the startup release check and `KONTEXT_NO_UPDATE_CHECK` in user-facing docs before shipping this. For an enterprise CLI, background outbound calls should be explicit.

Priority:
- Must-fix before merge: lifecycle/exit handling for the update checker.
- Should-fix soon: API version header, real fetch-path test coverage, and docs note for the outbound check/opt-out.

Final recommendation:
Block – must address the lifecycle issue before merge.

cmd/kontext/main.go
```diff
@@ -52,14 +53,17 @@ func startCmd() *cobra.Command {
...
+			return err
```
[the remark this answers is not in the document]
tumberger: Fixed in 115f6f9. Removed the `os.Exit` from `run.Start` entirely and moved exit-code propagation to `main.go`, after the updater fires. `CheckAsync` is now fully fire-and-forget (no channel, no wait), so the update check never blocks the command path regardless of how `run.Start` exits.

cmd/kontext/main.go
tumberger: Fixed in 115f6f9. Agreed — removed the wait entirely. `CheckAsync` is now `func CheckAsync(currentVersion string)` with no return value. The `os.Exit` that was in `run.Start` has been moved to `main.go` after the `run.Start` call, so the updater goroutine is never in the exit path. Fully fire-and-forget.

internal/update/update.go
tumberger: Done in 115f6f9. Added `X-GitHub-Api-Version: 2022-11-28` header. Also added a test assertion that verifies the header is set.

internal/update/update_test.go
tumberger: Refactored in 115f6f9. Extracted `githubAPIBase` as a package-level var that tests override to point at an `httptest.Server`. `TestFetchLatest` now calls the real `fetchLatest()` function and verifies headers, response parsing, and the normalised version. Also added `TestFetchLatestNon200` and `TestFetchLatestMalformedJSON` for error paths.

internal/update/update.go
tumberger: Correction: updated to `2026-03-10` in b541fd6 — that's the actual latest GitHub REST API version. `2022-11-28` was the legacy default.

michiosw (approved):
Re-reviewed the latest head. The previous blockers are addressed: exit-code propagation now lives in `main`, the updater is truly fire-and-forget, the GitHub API version is pinned, the fetch path is tested directly, and the README now documents the outbound check and opt-out env var. I reran `go test ./...` and `go test -race ./internal/update/...` on the updated branch and both passed.

PR #88, opened 2026-05-01
fix: add exponential backoff to sidecar heartbeat loop

### LINKED ISSUE #38: fix: sidecar heartbeat spams repeated errors without backoff
## Problem

The sidecar heartbeat loop runs on a fixed 30s ticker and logs every failure with `log.Printf`:

```go
if err := s.client.Heartbeat(ctx, s.sessionID); err != nil {
    log.Printf("sidecar: heartbeat: %v", err)
}
```

When the network goes down (DNS failure, WiFi switch, laptop sleep/wake), this produces the same error message every 30 seconds for the rest of the session:

```
2026/04/09 10:41:17 sidecar: heartbeat: unavailable: token refresh: token expired and refresh failed: oauth discovery: discovery request: Get "https://api.kontext.security/.well-known/oauth-authorization-server": dial tcp: lookup api.kontext.security: no such host
2026/04/09 10:41:47 sidecar: heartbeat: unavailable: ...
2026/04/09 10:42:17 sidecar: heartbeat: unavailable: ...
```

Same issue applies to `ingestEvent` failures at line 122-124.

## Expected behavior

1. **Exponential backoff** on consecutive failures (30s → 60s → 120s → cap at 5min)
2. **Deduplicate logs** — print the error on first occurrence and when it changes, not every tick
3. **Log recovery** — print a success message when connectivity returns after a failure streak
4. **Reset backoff** on success — return to the normal 30s interval once a heartbeat succeeds

## Example output

```
sidecar: heartbeat failed: dial tcp: lookup api.kontext.security: no such host (retrying with backoff)
sidecar: heartbeat recovered after 3m20s
```

KuaaMU:
## Summary

Add exponential backoff to sidecar heartbeat loop and deduplicate error logs.

## Motivation

Fixes #38. When the network goes down, the heartbeat logs the same error every 30 seconds indefinitely. This spams the logs and makes debugging harder.

## Changes

- Replace fixed 30s ticker with exponential backoff on consecutive failures (30s → 60s → 120s → 240s → cap at 5min)
- Deduplicate error logs: only log on first occurrence and when the error message changes
- Log recovery: print "heartbeat recovered after Xm Ys" when connectivity returns
- Reset backoff on success: return to 30s interval
- Extract `client` interface for testability (satisfied implicitly by `*backend.Client`)
- Add 3 tests: deduplication, different errors, backoff calculation

## Verification

```
go test ./internal/sidecar/... -v
go vet ./...
go build ./internal/run/...
```

All 8 tests pass, vet clean, callers still compile.


michiosw:
Thanks for this, really appreciate the first external contribution here.

The issue was valid: the sidecar heartbeat could spam the same network error every 30s, which is noisy and not useful when someone is offline or the API is temporarily unreachable. The backoff + deduping approach was the right direction, and it keeps the CLI quieter while still surfacing real failures and recovery.

I did a small cleanup pass before merging so the tests exercise the actual heartbeat state machine instead of duplicating the logic in the test, then let CI run green. This shipped in v0.5.1.

Thanks again for jumping in here — genuinely cool to see this come from outside the core team.

PR #92, opened 2026-05-02
feat: add hosted access enforcement

### LINKED ISSUE #91: Unify hook adapter runtime across Guard and hosted mode
## Problem

We are about to add Cursor hook support, but the CLI currently has two separate hook paths:

- Hosted mode: `kontext start --agent claude` generates hooks that call `kontext hook --agent claude`, then events go through `internal/agent` / `internal/hook` into the sidecar and hosted backend.
- Guard mode: `kontext guard start` installs hooks that call `kontext guard hook claude-code`, then events go through the Guard hook runtime into the local daemon and risk engine.

PR #90 refactors the Guard hook runtime, but that adapter boundary is still Guard-specific. If we add Cursor only there, we will solve local Guard while likely duplicating Cursor protocol handling later for hosted mode.

## Why this matters

Cursor support should not require two implementations of the same agent hook protocol. The agent-specific adapter should understand Claude or Cursor hook JSON once, then route the normalized action to whichever runtime is active.

## Proposed architecture

Create one shared hook adapter/runtime layer:

```text
agent hook protocol adapters
  Claude adapter
  Cursor adapter

shared normalized event/action model

runtimes / sinks
  hosted sink -> sidecar -> backend ProcessHookEvent
  guard sink  -> local daemon -> risk decision
```

Guard and hosted mode should share the same protocol adapters and normalized event model. They should differ only in the processor/sink:

- Guard processor: local daemon, local risk decision, local SQLite/dashboard.
- Hosted processor: sidecar, hosted `ProcessHookEvent`, governed session traces.

## Suggested scope

- Move the hook runtime out of `internal/guard` into a shared package.
- Replace Guard-specific `risk.HookEvent` at the runtime boundary with a neutral hook/action event type.
- Add a Guard policy/processor adapter that maps neutral events into the existing risk model.
- Merge or delete the older `internal/agent` / `internal/hook` abstraction so there is one hook runtime path.
- Keep install/config code agent-specific, because Claude settings and Cursor `.cursor/hooks.json` are different surfaces.

## Acceptance criteria

- Claude hook decoding/encoding is implemented once and works for both Guard and hosted mode.
- Cursor support can be added as one adapter and reused by both Guard and hosted mode.
- Guard and hosted mode choose different processors/sinks without duplicating agent protocol logic.
- The implementation follows the repo's hard-cutover approach: no long-term parallel compatibility layer.

michiosw:
## Summary

- Implements the hosted Claude hook runtime path for Kontext CLI and closes the shared-hook extraction work for #91.
- Adds first-class generated proto fields for hosted access mode, policy epoch, reason code, and request id.
- Reuses one Claude hook adapter for hosted and local Guard mode, preserving Guard behavior while moving decode/encode logic into `internal/hookruntime`.
- Makes hosted `PreToolUse` synchronous through the sidecar, with enforce-mode fail-closed behavior and disabled/no-policy fail-open behavior.
- Adds sanitized GitHub git context enrichment for Bash hooks and verifies blocking hook settings before launching an enforcing hosted session.
- Gates managed GitHub credential delivery on backend ALLOW by returning Claude `updatedInput` for the current Bash command only; ASK/DENY never resolve or inject managed credentials.

Closes #91.

## Verification

- [x] `buf generate`
- [x] `go test ./...`
- [x] `git diff --check`

## Notes

- Depends on `kontext-security/proto#8` for the response metadata contract.
- In enforce mode, managed provider credentials are not preloaded into the Claude process. For allowed GitHub Bash tool calls, the sidecar exchanges the managed credential after the backend ALLOW decision and returns a command-scoped `updatedInput` that reads the token from a session-local 0600 file, so raw token material is not emitted in hook JSON.

## Release impact

- Minor: `feat:` hosted access enforcement path and generated protocol fields.

### DIFF
--- a/buf.gen.yaml
+++ b/buf.gen.yaml
@@ -6,8 +6,8 @@ managed:
       value: github.com/kontext-security/kontext-cli/gen
 inputs:
   - git_repo: https://github.com/kontext-security/proto.git
-    branch: main
-    ref: a9da139386c97cb653497ea08ae84fe27dffc850
+    branch: feat/hosted-access-response-metadata
+    ref: 10b65e8a783e3462596d233254bb43cc4bc7b862
 plugins:
   - local: protoc-gen-go
     out: gen
--- a/cmd/kontext/main.go
+++ b/cmd/kontext/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"net"
@@ -14,8 +13,10 @@ import (
 
 	"github.com/kontext-security/kontext-cli/internal/agent"
 	"github.com/kontext-security/kontext-cli/internal/auth"
+	"github.com/kontext-security/kontext-cli/internal/backend"
 	guardcli "github.com/kontext-security/kontext-cli/internal/guard/cli"
 	"github.com/kontext-security/kontext-cli/internal/hook"
+	"github.com/kontext-security/kontext-cli/internal/hookruntime"
 	"github.com/kontext-security/kontext-cli/internal/run"
 	"github.com/kontext-security/kontext-cli/internal/sidecar"
 	"github.com/kontext-security/kontext-cli/internal/update"
@@ -92,7 +93,7 @@ func startCmd() *cobra.Command {
 			err := run.Start(ctx, run.Options{
 				Agent:        agentName,
 				TemplateFile: templateFile,
-				IssuerURL:    auth.DefaultIssuerURL,
+				IssuerURL:    backend.BaseURL(),
 				ClientID:     auth.DefaultClientID,
 				Verbose:      verbose,
 				Args:         args,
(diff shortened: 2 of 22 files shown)


michiosw:
* **#92** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/92?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/92?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tumberger:
@michiosw quick scope check on #91 before this lands:

This PR looks like it solves the most important part of #91: Claude hook decode/encode moves into shared `internal/hookruntime`, and both hosted and Guard reuse that codec. That seems sufficient to avoid duplicating Claude/Cursor protocol handling.

One thing still seems outside this PR, though: we still have two runtime paths:
- hosted: `internal/agent` + `internal/hook` + sidecar
- Guard: `internal/guard/hookruntime` + local daemon/risk

And the Guard runtime boundary still uses `risk.HookEvent` / `risk.Decision`, while `internal/agent` and `internal/hook` remain.

Should we treat #91 as closed by the shared protocol codec work, or keep a follow-up for the harder cutover to one neutral runtime/sink abstraction? Given the repo hard-cutover approach, I would lean toward tracking that explicitly if it will not be included here.

michiosw:
> @michiosw quick scope check on #91 before this lands:
>
> This PR looks like it solves the most important part of #91: Claude hook decode/encode moves into shared `internal/hookruntime`, and both hosted and Guard reuse that codec. That seems sufficient to avoid duplicating Claude/Cursor protocol handling.
>
> One thing still seems outside this PR, though: we still have two runtime paths:
>
> - hosted: `internal/agent` + `internal/hook` + sidecar
> - Guard: `internal/guard/hookruntime` + local daemon/risk
>
> And the Guard runtime boundary still uses `risk.HookEvent` / `risk.Decision`, while `internal/agent` and `internal/hook` remain.
>
> Should we treat #91 as closed by the shared protocol codec work, or keep a follow-up for the harder cutover to one neutral runtime/sink abstraction? Given the repo hard-cutover approach, I would lean toward tracking that explicitly if it will not be included here.

​thanks for the comments, will add you as reviewer to review once it's ready - as usual review bots only trigger if pr is on "open" not draft mode hence why I put it not in draft - i am still working on it

michiosw:
Superseded by the reviewable Graphite stack: #93 (shared Claude hook runtime) and #94 (hosted hook sidecar transport).

PR #199, opened 2026-05-22

michiosw:
## Where We Are

Installing or uninstalling Kontext Guard Claude hooks can delete unrelated hooks when they share the same Claude hook group. The current code removes whole groups based on a string match, so mixed groups lose non-Guard commands.

## Where We Want To Go

Guard hook management should remove only Kontext Guard commands and leave unrelated Claude hooks and their matchers intact. That keeps hook install and uninstall safe for users who already have custom commands in the same group.

## How do we get there

This change filters nested hook commands instead of dropping whole groups, keeps mixed groups when non-Guard commands remain, and adds regression coverage for both merge and uninstall paths in `internal/guard/cli`. Verified with `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check`.

### DIFF
--- a/internal/guard/cli/cli.go
+++ b/internal/guard/cli/cli.go
@@ -406,8 +406,8 @@ func uninstallClaudeHooks(out io.Writer) error {
 		}
 		filtered := make([]any, 0, len(list))
 		for _, entry := range list {
-			if !isGuardHookEntry(entry) {
-				filtered = append(filtered, entry)
+			if updated, keep := removeGuardHookCommands(entry); keep {
+				filtered = append(filtered, updated)
 			}
 		}
 		hooks[hookName] = filtered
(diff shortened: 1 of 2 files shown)


michiosw:
* **#199** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/199?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/199?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

hasandemirkiran:
### Merge activity

* **May 22, 9:52 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/199).
* **May 22, 9:53 AM UTC**: @hasandemirkiran merged this pull request with [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/199).

PR #202, opened 2026-05-23
test(judge): deflake llama-server process timeouts

michiosw:
Summary
This cleans up internal/guard/judge llama-server process tests by making the startup and early-exit timeouts resilient under parallel and race load.

Before this, go test ./... and go test -race ./... could fail on main because the llama-server harness used a 2s startup timeout and a 1.5s early-exit threshold, which was too tight when tests run concurrently.

Now the tests use shared, more realistic thresholds (15s startup timeout; 5s early-exit max), so they pass reliably without changing runtime behavior.

Why
This gives kontext-cli a cleaner maintenance path for the judge runtime test harness:

StartLlamaServer(...)
-> waitHealthy(...) polling loop
-> fake llama-server subprocess
-> stable tests under parallel/race execution

This PR does not broaden behavior beyond the cleanup scope.

What changed
Added shared llama-server test timeout constants
Relaxed startup timeout to 15s
Relaxed early-exit assertion to 5s max

Verification
go test ./internal/guard/judge -count=1
go test ./... -count=1
go vet ./...
go test -race ./...
git diff --check

### DIFF
--- a/internal/guard/judge/llama_server_test.go
+++ b/internal/guard/judge/llama_server_test.go
@@ -15,6 +15,11 @@ import (
 	"time"
 )
 
+const (
+	llamaServerTestStartupTimeout      = 15 * time.Second
+	llamaServerTestEarlyExitMaxElapsed = 5 * time.Second
+)
+
 func TestBuildLlamaServerArgsUsesLocalModel(t *testing.T) {
 	got := BuildLlamaServerArgs(LlamaServerOptions{
 		ModelPath: "/models/qwen.gguf",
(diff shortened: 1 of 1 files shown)


michiosw:
* **#202** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/202?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/202?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #215, opened 2026-05-24

michiosw:
## Where We Are

Clawpatch found that `runtimecore` accepts unknown hook names and routes them through the telemetry path instead of rejecting them. That lets malformed or case-shifted enforcement events bypass fail-closed validation.

## Where We Want To Go

Unknown hook names should fail immediately before the runtime decides whether to evaluate or ingest them. That keeps enforcement strict and makes bad hook payloads visible instead of silently downgrading them.

## How do we get there

Added `HookName.IsKnown()` in `internal/hook/domain.go` and used it in both `ValidateEvaluateHook` and `ValidateIngestEvent` so unknown hook names return an error before routing. Added a regression test that proves `ProcessHook` rejects an unknown hook without calling either runtime path. Verified with `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check`.

### DIFF
--- a/internal/hook/domain.go
+++ b/internal/hook/domain.go
@@ -20,6 +20,15 @@ func (h HookName) String() string {
 	return string(h)
 }
 
+func (h HookName) IsKnown() bool {
+	switch h {
+	case HookSessionStart, HookPreToolUse, HookPostToolUse, HookPostToolUseFailed, HookSessionEnd, HookUserPromptSubmit:
+		return true
+	default:
+		return false
+	}
+}
+
 func (h HookName) CanBlock() bool {
 	return h == HookPreToolUse
 }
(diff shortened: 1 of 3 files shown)


michiosw:
* **#215** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/215?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/215?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #222, opened 2026-05-26

michiosw:
## Where We Are

The guard dashboard can show the wrong active policy after a successful activation because an older `fetchPolicy()` response can land after `activatePolicy()` and overwrite the newer state. The first PR attempt also failed CI because the regenerated Guard dashboard assets were not checked in.

## Where We Want To Go

Policy activation should win over any stale background load so the dashboard always reflects the latest active policy, and the repo should include the matching built dashboard assets so CI stays green.

## How do we get there

This change tags policy loads with a request id plus the current mutation version, and ignores any load response that started before a policy activation. It also clears stale policy errors after a successful activation and checks in the rebuilt Guard dashboard assets that correspond to the dashboard change. Local validation passed for `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `./scripts/build-guard-dashboard-assets.sh`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check`.


michiosw:
* **#222** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/222?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/222?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #226, opened 2026-05-27

michiosw:
## Where We Are

The Local Guard daemon binds an unauthenticated Unix socket under a predictable /tmp path, but it accepted any pre-existing parent directory state there. That left a hole where a symlinked, permissive, or wrong-owner socket directory could weaken the local trust boundary before the daemon started.

## Where We Want To Go

We want Guard startup to fail closed unless the socket parent directory is a real directory owned by the current user with 0700 permissions, so the runtime socket stays private to the local user.

## How do we get there

This change hardens localruntime.EnsureSocketDir to mirror the managed-observe checks: create the directory if needed, reject symlinks and non-directories, verify current-user ownership, and chmod the directory to 0700 when needed. It also adds focused tests that prove an existing permissive directory is tightened and a symlink parent is rejected. Verification: `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check` all passed.

### DIFF
--- a/internal/localruntime/socket.go
+++ b/internal/localruntime/socket.go
@@ -1,9 +1,11 @@
 package localruntime
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+	"syscall"
 )
 
 func DefaultSocketPath() string {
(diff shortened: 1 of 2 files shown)


michiosw:
* **#226** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/226?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/226?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #230, opened 2026-05-28
fix(clawpatch): guard local runtime socket cleanup

michiosw:
## Where We Are

Starting the local runtime with a custom `--socket` path can delete an arbitrary existing file before the daemon binds its Unix socket. A mistyped socket path can destroy user data instead of failing safely.

## Where We Want To Go

Starting the local runtime should only remove a stale Unix socket. Any regular file, directory, symlink, or other non-socket path should stay intact and return a clear error.

## How do we get there

Check the existing path with `os.Lstat` before cleanup, remove it only when it is a Unix socket, and return an error for any other file type. Added a regression test that pre-creates a regular file at the socket path and asserts `Start` fails without deleting it. Verified with `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check`.

### DIFF
--- a/internal/localruntime/service.go
+++ b/internal/localruntime/service.go
@@ -64,8 +64,16 @@ func NewService(opts Options) (*Service, error) {
 func (s *Service) SocketPath() string { return s.socketPath }
 
 func (s *Service) Start(ctx context.Context) error {
-	if err := os.Remove(s.socketPath); err != nil && !errors.Is(err, os.ErrNotExist) {
-		return fmt.Errorf("remove stale local runtime socket: %w", err)
+	info, err := os.Lstat(s.socketPath)
+	if err == nil {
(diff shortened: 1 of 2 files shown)


michiosw:
* **#230** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/230?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/230?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #234, opened 2026-05-30

michiosw:
## Where We Are

The root workspace test script skipped the documented `go test -race ./...` check. A change could pass `pnpm test` while still missing the race detector run that the README tells contributors to use.

## Where We Want To Go

`pnpm test` should enforce the same race detector check the repo documents, so the default local test path catches data races before merge.

## How do we get there

Update the root `test` script in `package.json` to run `go test -race ./...` before `go vet ./...`. Verification: `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check` all passed in the temp checkout.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "packageManager": "pnpm@10.0.0",
   "scripts": {
     "build": "./scripts/build-guard-dashboard-assets.sh",
-    "test": "go test ./... && go vet ./...",
+    "test": "go test -race ./... && go vet ./...",
     "dev": "pnpm --dir web/guard-dashboard dev"
   }
 }


michiosw:
* **#234** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/234?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/234?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #240, opened 2026-06-01

michiosw:
## Where We Are

The guard dashboard can show the wrong error after a session switch. If session A starts loading, the user switches to session B, and A fails later, A's error still replaces the visible state while B is selected.

## Where We Want To Go

Only the selected session should be allowed to update the dashboard state. A stale failure from an old session must be ignored so the error banner still matches the session on screen.

## How do we get there

This change adds the same `selectedRef.current !== id` guard to the `loadEvents` error path that already exists on the success path in `web/guard-dashboard/src/App.tsx`. Validation passed with `go test ./...` on rerun after one unrelated flaky timeout in `internal/guard/judge`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check`.


michiosw:
* **#240** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/240?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/240?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #243, opened 2026-06-02
[AUTOMATION] fix(clawpatch): preserve claude hook file permissions

michiosw:
## Where We Are

Installing or uninstalling Claude guard hooks rewrites `~/.claude/settings.json` and its backup with broad permissions. A user can start from `0600` and end up with readable files for other local users.

## Where We Want To Go

Keep the Claude settings file and backup private when guard hooks update them. Existing restrictive modes should stay in place, and new writes should default to `0600`.

## How do we get there

Preserve the current file mode in `backupFile` and `writeJSONFile`, with a `0600` fallback for new files. Add a regression test that seeds `settings.json` at `0600`, runs hook install, and checks both the rewritten settings file and backup mode. Verified with `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check`.

### DIFF
--- a/internal/guard/cli/cli.go
+++ b/internal/guard/cli/cli.go
@@ -440,7 +440,8 @@ func readClaudeSettings() (string, map[string]any, error) {
 }
 
 func backupFile(path, label string) error {
-	if _, err := os.Stat(path); os.IsNotExist(err) {
+	info, err := os.Stat(path)
+	if os.IsNotExist(err) {
 		return nil
 	} else if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


michiosw:
* **#243** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/243?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/243?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #246, opened 2026-06-03
fix(hook): decode array tool_response from MCP tools

hasandemirkiran:
## Problem
The `PostToolUse` hook couldn't decode **any** MCP tool result. `claudeHookInput.tool_response` was typed `map[string]any`, but Claude Code delivers MCP tool results (Linear, Gmail, etc.) as a JSON **array** of content blocks. `json.Unmarshal` rejected the whole payload, so the hook aborted before recording the event.

**Impact:** every MCP tool call was silently skipping observation. Built-in tools (which return objects) were unaffected.

## Fix
`internal/hookruntime/claude.go`:
- Changed `tool_response` fields to `json.RawMessage` so decoding no longer fails on shape.
- Added `normalizeToolResponse()`: object payloads pass through unchanged; arrays (and any other non-object shape) are wrapped under `"content"`, keeping the downstream `map[string]any` pipeline (observe DB hashing, risk types) unchanged. Nothing reads specific response keys, so the wrapper is safe.
- Added regression tests for the object / array / absent cases.

## Verification
- `go build ./...` clean
- `go test ./internal/hookruntime/... ./internal/localruntime/...` passes
- Rebuilt `bin/kontext` and confirmed a live Linear MCP call no longer errors

Linear: ENG-430 — https://linear.app/cobrowser/issue/ENG-430/fix-posttooluse-hook-dropping-mcp-tool-responses-array-tool-response

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/kontext-security/kontext-cli
 
-go 1.25.10
+go 1.25.11
 
 require (
 	connectrpc.com/connect v1.19.2
--- a/internal/hookruntime/claude.go
+++ b/internal/hookruntime/claude.go
@@ -1,6 +1,7 @@
 package hookruntime
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"strings"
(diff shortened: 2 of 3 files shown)


michiosw (dismissed):

hasandemirkiran:
@greptileai

michiosw (approved):

michiosw:
### Merge activity

* **Jun 5, 9:51 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/246).
* **Jun 5, 9:51 AM UTC**: @michiosw merged this pull request with [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/246).

PR #258, opened 2026-06-07

michiosw:
## Where We Are

`kontext guard start` installed Claude hooks before the daemon proved it could bind and serve. If startup failed after that write, Claude was left pointing at a dead local Guard hook.

## Where We Want To Go

Daemon startup should only install Claude hooks after the local runtime and HTTP listener are ready. A failed start should leave `~/.claude/settings.json` unchanged.

## How do we get there

Bind the TCP listener before hook installation, then serve with that pre-opened listener. Add a regression test that starts the daemon with an invalid listen address and proves no Claude settings file is created. Verified with `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check`.

### DIFF
--- a/internal/guard/cli/cli.go
+++ b/internal/guard/cli/cli.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


michiosw:
* **#258** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/258?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/258?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #261, opened 2026-06-08

michiosw:
## Where We Are

Policy activation in the guard dashboard can race. Two clicks in the same tick can start two requests before React commits `policyPending`, and the older response can overwrite the newer selection.

## Where We Want To Go

Only one policy activation request should be allowed at a time. The dashboard should keep the active profile in sync with the user's latest accepted click.

## How do we get there

Use a synchronous ref guard in `web/guard-dashboard/src/App.tsx` so the second same-tick activation returns before another request starts. Only apply the response and clear the pending state when the finishing request still matches that ref.

Validation:
- `go vet ./...`
- `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`
- `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`
- `git diff --check`
- `go test ./...` still fails in baseline `internal/guard/judge` timing tests `TestStartLlamaServerHealthCheckAndStop` and `TestStartLlamaServerEarlyExitDoesNotWaitForStopTimeout`


michiosw:
* **#261** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/261?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/261?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #266, opened 2026-06-08
fix(managedstream): add stored timestamp cursor keys

michiosw:
## Where We Are

Katana managed observe can still fail hosted ingest when legacy ledger rows contain timestamps without a timezone, like 2026-06-08T12:20:07.853885. Hosted expects offset-aware timestamps, so those rows turn into 400 responses.

## Where We Want To Go

Managed observe should upload the same ledger rows with valid UTC timestamps and keep cursor pagination stable. It must not skip rows, leak internal cursor fields into the payload, or rely on SQLite timestamp rounding.

## How do we get there

This adds a stored Go-computed updated_at cursor key, backfills it during SQLite migration, uses it for bounded cursor filtering, and normalizes exported *_at fields to RFC3339 UTC. Covered by focused ledger/managed-stream tests and verified with go test -count=1 ./internal/guard/store/sqlite ./internal/managedstream plus go test ./...

### DIFF
--- a/internal/guard/store/sqlite/ledger.go
+++ b/internal/guard/store/sqlite/ledger.go
@@ -12,9 +12,10 @@ import (
 type LedgerRecord map[string]any
 
 type LedgerExportOptions struct {
-	UpdatedAfter   *time.Time
-	UpdatedAfterID string
-	Limit          int
+	UpdatedAfter    *time.Time
+	UpdatedAfterKey string
+	UpdatedAfterID  string
+	Limit           int
 }
 
 type LedgerBatch struct {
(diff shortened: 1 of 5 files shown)


michiosw:
* **#266** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/266?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/266?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

michiosw:
@greptileai review

PR #267, opened 2026-06-08
fix(managedstream): use fixed ledger timestamp cursors

michiosw:
## Where We Are

Katana managed observe can fail hosted ingest when old local ledger rows contain timestamps without offsets, exact-second timestamps, or corrupt timestamp strings. The old export path used raw timestamp text for both hosted payloads and cursor pagination.

## Where We Want To Go

Managed observe should keep uploading valid ledger batches and keep cursor pagination stable. It should not skip same-second rows, upload invalid datetime strings, leak internal cursor fields, or fail store migration because one legacy timestamp is bad.

## How do we get there

This adds a fixed-width UTC cursor key for authorization actions, backfills it during SQLite migration, and uses it for bounded cursor queries. Exported ledger timestamps are normalized to hosted-valid RFC3339 strings; corrupt optional timestamps become null, and corrupt required timestamps fall back to a valid UTC timestamp. Managed-stream state parsing now accepts old no-offset and SQLite-space timestamp formats. Verified with go test -count=1 ./internal/guard/store/sqlite ./internal/managedstream, go test ./..., and pnpm --dir /Users/michelosswald/cobro/code/workspace/kontext --filter api test -- authorization-ledger.validation.spec.ts --runInBand.

### DIFF
--- a/internal/guard/store/sqlite/ledger.go
+++ b/internal/guard/store/sqlite/ledger.go
@@ -34,6 +34,13 @@ type LedgerCursor struct {
 	ActionID  string
 }
 
+const (
+	ledgerCursorUpdatedAtKeyColumn = "__cursor_updated_at_key"
+	ledgerTimestampCursorKeyLayout = "2006-01-02T15:04:05.000000000Z07:00"
+)
+
+var ledgerFallbackTimestamp = time.Unix(0, 0).UTC()
+
 const authorizationActionsSelect = `
 select id, session_id, turn_id, tool_use_id, trace_id, span_id, parent_span_id,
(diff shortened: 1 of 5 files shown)


michiosw:
* **#267** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/267?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/267?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #270, opened 2026-06-09

michiosw:
## Where We Are

The local Guard runtime accepts any existing socket parent directory under `/tmp` as long as `os.MkdirAll` succeeds. If that directory is a symlink or belongs to another uid, the daemon can remove and bind `kontext.sock` inside an untrusted path.

## Where We Want To Go

The local Guard runtime should only bind its socket inside a real directory owned by the current user, with `0700` permissions. That keeps the Guard socket path under the same ownership rules as managed observe.

## How do we get there

Mirror the managed-observe socket-directory checks in `internal/localruntime.EnsureSocketDir`: reject symlink parents, reject non-directories, require the current uid, and tighten permissions back to `0700` when needed. Add focused localruntime tests for writable-directory tightening, non-directory rejection, and symlink rejection. Verified with `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check` from the isolated checkout.

### DIFF
--- a/internal/localruntime/socket.go
+++ b/internal/localruntime/socket.go
@@ -1,9 +1,11 @@
 package localruntime
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+	"syscall"
 )
 
 func DefaultSocketPath() string {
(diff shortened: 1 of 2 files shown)


michiosw:
* **#270** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/270?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/270?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #273, opened 2026-06-10

michiosw:
## Where We Are

The local Guard daemon keeps running after the caller cancels context. `runDaemon` started the runtime service with `context.Background()` and the HTTP server had no shutdown path tied to the caller context, so tests or supervisors could hang until process exit.

## Where We Want To Go

The daemon should stop when the caller context is canceled. `kontext start` should return cleanly after shutdown instead of leaving the socket and HTTP listener alive.

## How do we get there

Thread the caller context into the local runtime service and the HTTP server, treat `http.ErrServerClosed` as a normal shutdown, and add a regression test that starts the daemon, waits for it to come up, cancels context, and asserts clean return. Verified with `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check`.

### DIFF
--- a/internal/guard/app/server/server.go
+++ b/internal/guard/app/server/server.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/fs"
 	"mime"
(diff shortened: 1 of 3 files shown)


michiosw:
* **#273** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/273?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/273?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #275, opened 2026-06-11
[AUTOMATION] fix(clawpatch): refuse non-socket local runtime paths

michiosw:
## Where We Are

`kontext guard start --socket <path>` deletes any existing file at that path before it checks whether the path is actually a Unix socket. A user can point Guard at `~/notes.txt` and lose that file.

## Where We Want To Go

Guard should only remove stale Unix sockets. If the configured path is a regular file, directory, or anything else, startup should stop and leave that path untouched.

## How do we get there

The local runtime now uses `os.Lstat` before unlinking the socket path and returns an error unless the existing path is a socket. A regression test writes a regular file at the configured socket path, starts the service, and verifies startup fails without changing the file contents.

Validation passed with `go test ./internal/guard/judge`, `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check`.

### DIFF
--- a/internal/localruntime/service.go
+++ b/internal/localruntime/service.go
@@ -64,8 +64,8 @@ func NewService(opts Options) (*Service, error) {
 func (s *Service) SocketPath() string { return s.socketPath }
 
 func (s *Service) Start(ctx context.Context) error {
-	if err := os.Remove(s.socketPath); err != nil && !errors.Is(err, os.ErrNotExist) {
-		return fmt.Errorf("remove stale local runtime socket: %w", err)
+	if err := removeStaleSocket(s.socketPath); err != nil {
+		return err
 	}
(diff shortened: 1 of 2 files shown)


michiosw:
* **#275** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/275?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/275?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #279, opened 2026-06-11
feat(githubpolicy): evaluate synced GitHub policy locally in observer mode (ENG-426)

hasandemirkiran:
## Summary

Consumer side of ENG-426: managed endpoints now fetch the cloud GitHub access-control policy snapshot (`github-policy-snapshot-v1`, served by kontext PR #551), cache it per installation, evaluate classified GitHub activity against it locally, and record would-allow/would-deny decisions — without ever blocking anything (observer mode).

Local decision layering: **general deterministic guardrails > synced policy > probabilistic signals** (`RiskPolicyProvider.DecideHook` evaluates the snapshot after the deterministic policy engine and before the judge; the enforce path is implemented but only an explicit `mode: "enforce"` can ever block — anything else is observe).

## What's in here

- **`internal/githubpolicy`** (new): vendored snapshot types mirroring `packages/shared/src/github-policy-snapshot.ts`; HTTP client for `GET /api/v1/policy/github/snapshot` (same install token as ledger ingest, no org parameter); per-installation disk cache + 60s refresh loop (`KONTEXT_GITHUB_POLICY_REFRESH_INTERVAL` to override); Go port of the cloud classifier (`provider-action-classifier.ts`); local evaluator mirroring the cloud semantics exactly (deny beats allow; org layer strict; user/agent layers abstain while empty, veto once active; exact matching, null = wildcard, no globs).
- **Ledger**: each evaluated action gets its own `request.decided` row with `decision_category: "dry_run"`, `reason_code` `ALLOWED_POLICY_CHECK`/`DENY_POLICY_CHECK`, `matched_rules_json` (rule ids + which one decided), `policy_version` = snapshot epoch, `policy_hash` = snapshot hash, `provider: github`, `resource_class: repo`, `resource_id: owner/name`, and `context_json.github.{owner,repo,branch_or_ref}` + `context_json.github_policy.{mode,epoch,hash,stale,subjects_resolved}`. Rows flow through the existing `POST /api/v1/authorization-ledger/batches` pipeline with chained receipts. Existing `request.proposed`/`observed` rows are untouched.
- **Cache resilience**: on fetch failure the last cached snapshot keeps evaluating (stale-but-deterministic beats unavailable) and staleness is recorded on every decision; unchanged `hash` skips re-applying; the persisted snapshot survives daemon restarts and starts out marked stale until the cloud confirms it.

## Identity decision (pilot caveat)

Per `docs/guides/access-control-event-schema.md`, the managed endpoint's trusted identity is the **service account + installation** — Claude hook payloads are not trusted human identity, and this repo has no enrolled Kontext user/application identity. So requests are evaluated **org-layer-first with unresolved user/agent subjects**: user/agent-layer rules never match their subject (unresolved subject = no match), which means a snapshot containing any user/agent-layer rule makes that layer active and silence vetoes — exactly what the cloud evaluator does for an unknown user id. Every dry-run row flags this via `context_json.github_policy.subjects_resolved: false`, and `identity_context_json.principal_kind` is `service_account`.

## Deliberate local extensions over the cloud classifier

- `gh` commands without `-R/--repo` fall back to the cwd's GitHub origin remote for the policy anchor (the task spec lists "current branch + remotes" as a derivation source; the cloud classifier only uses `-R` because it can't see the endpoint's cwd).
- `gh api --input …` counts as mutating, and a method override that can't be proven `GET/HEAD` defaults to write ("default to write when unprovable").

## Verification (cloud repo on PR #551 branch, localhost:4000, katana-local profile)

1. `curl` with the install token returns the snapshot; missing token → 401. Rule edits changed `epoch 6 / 60593dfb…` → `epoch 7 / a3e06c34…` and the daemon picked the change up on its own refresh tick (no restart).
2. Real `PreToolUse` hook events through the launchd daemon (`git push …enterprise-deployments.git main`, `gh pr create -R kontext-security/kontext`) produced `request.decided` `dry_run` rows in local `guard.db` **and** in `hosted_ledger_actions` via the existing stream.
3. Semantics table verified live: org allow alone ⇒ org-wide would-allow (`ALLOWED_POLICY_CHECK`); org allow + targeted deny ⇒ would-deny only on `kontext-security/enterprise-deployments` with the deny rule id marked as deciding; user-layer rule present ⇒ user layer strict (would-deny `no matching allow rule in active layer(s): user` with unresolved subject).
4. Every hook response stayed `permissionDecision: allow` — nothing was blocked, regardless of the would-be decision.

Unit tests cover the evaluator semantics table, classifier parity cases (incl. chained/wrapped commands and the `gh api` heuristic), cache staleness/refresh, the dry-run ledger rows + receipts, and the DecideHook layering (observe never blocks, unknown modes treated as observe, enforce reserved path denies before the judge).

### DIFF
--- a/internal/githubpolicy/classifier.go
+++ b/internal/githubpolicy/classifier.go
@@ -1,6 +1,7 @@
 package githubpolicy
 
 import (
+	"path/filepath"
 	"regexp"
 	"strings"
 )
@@ -57,39 +58,61 @@ var githubHostRE = regexp.MustCompile(`(?i)(^|[\s"'(<])((https?://)?(api\.)?gith
 // branch resolution — because deriving it shells out to git.
 func ClassifyProviderActions(toolName string, toolInput map[string]any, gitContext func() GitContext) []ProviderAction {
 	if gitContext == nil {
-		gitContext = func() GitContext { return GitContext{} }
+		return ClassifyProviderActionsWithCWD(toolName, toolInput, "", nil)
 	}
+	return ClassifyProviderActionsWithCWD(toolName, toolInput, "", func(string) GitContext {
+		return gitContext()
+	})
+}
+
+// ClassifyProviderActionsWithCWD maps a tool invocation to canonical GitHub
+// actions, using cwd as the base for git -C context resolution.
+func ClassifyProviderActionsWithCWD(toolName string, toolInput map[string]any, cwd string, gitContext func(string) GitContext) []ProviderAction {
+	if gitContext == nil {
+		gitContext = func(string) GitContext { return GitContext{} }
+	}
+	loadGitContext := memoizedGitContext(gitContext)
 	switch strings.ToLower(toolName) {
 	case "bash", "shell":
-		command, _ := stringField(toolInput, "command")
-		if command == "" {
-			command, _ = stringField(toolInput, "cmd")
-		}
+		command := commandFromToolInput(toolInput)
 		if command == "" {
 			return nil
 		}
-		return classifyCommand(command, memoizedGitContext(gitContext))
+		return classifyCommand(command, cwd, loadGitContext)
 	case "webfetch", "web_fetch":
 		url, _ := stringField(toolInput, "url")
 		if url != "" && isGithubURL(url) {
 			return []ProviderAction{{Action: "github.api.read", Resource: normalizeGithubRepo(url)}}
 		}
 	}
+	if action, ok := classifyGithubMCPTool(toolName, toolInput); ok {
+		return []ProviderAction{action}
+	}
 	return nil
 }
 
-func memoizedGitContext(load func() GitContext) func() GitContext {
-	var cached *GitContext
-	return func() GitContext {
-		if cached == nil {
-			context := load()
-			cached = &context
+func commandFromToolInput(input map[string]any) string {
+	for _, key := range []string{"command", "cmd", "script"} {
+		if value, _ := stringField(input, key); value != "" {
+			return value
 		}
-		return *cached
+	}
+	return ""
+}
+
+func memoizedGitContext(load func(string) GitContext) func(string) GitContext {
+	cached := map[string]GitContext{}
+	return func(cwd string) GitContext {
+		if context, ok := cached[cwd]; ok {
+			return context
+		}
+		context := load(cwd)
+		cached[cwd] = context
+		return context
 	}
 }
 
-func classifyCommand(command string, gitContext func() GitContext) []ProviderAction {
+func classifyCommand(command string, cwd string, gitContext func(string) GitContext) []ProviderAction {
 	trimmed := strings.TrimSpace(command)
 	if trimmed == "" {
 		return nil
@@ -98,15 +121,17 @@ func classifyCommand(command string, gitContext func() GitContext) []ProviderAct
 	for _, segment := range splitCommandSegments(trimmed) {
 		invocation, args := commandInvocation(segment)
 		if unwrapped, ok := unwrapShellLauncher(invocation, args); ok {
-			actions = append(actions, classifyCommand(unwrapped, gitContext)...)
+			actions = append(actions, classifyCommand(unwrapped, cwd, gitContext)...)
 			continue
 		}
 		switch invocation {
 		case "gh":
-			actions = append(actions, classifyGhCommand(args, segment, gitContext))
+			actions = append(actions, classifyGhCommand(args, segment, func() GitContext {
+				return gitContext(cwd)
+			}))
 			continue
 		case "git":
-			actions = append(actions, classifyGitCommand(args, segment, gitContext)...)
+			actions = append(actions, classifyGitCommand(args, segment, cwd, gitContext)...)
 			continue
 		case "curl", "wget", "http", "https", "httpie":
 			if isGithubURL(segment) {
@@ -173,8 +198,10 @@ func classifyGhCommand(args []string, raw string, gitContext func() GitContext) 
 	case "workflow":
 		mutating := verb == "run" || verb == "enable" || verb == "disable"
 		return ProviderAction{Action: apiAction(mutating), Resource: resource}
+	case "release", "secret", "gist":
+		return ProviderAction{Action: readWriteAction("github.api", isReadVerb(verb)), Resource: resource}
 	}
-	return ProviderAction{Action: apiAction(commandLooksMutating(raw)), Resource: resource}
+	return ProviderAction{Action: apiAction(commandLooksMutating(raw) || !isKnownReadOnlyGhCommand(area, verb)), Resource: resource}
 }
 
 func readWriteAction(prefix string, read bool) string {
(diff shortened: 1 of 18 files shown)


hasandemirkiran:
Addressed all review findings in a26a24b:

- **gh api endpoint path** now anchors the resource (`gh api repos/acme/admin/issues` → `acme/admin`); `{owner}/{repo}` placeholders still resolve via `-R`/cwd.
- **gh repo positional argument** is honored (`gh repo delete acme/admin` no longer attributes to the cwd repo).
- **curl/wget/bare-URL and WebFetch** GitHub calls extract `owner/repo` from the URL, including `api.github.com/repos/...` paths (non-repos API paths intentionally carry no resource).
- **Implicit-remote `git push`/`pull`/`fetch`** now evaluate against the `origin` remote instead of producing no evaluation.
- **`git clone`** no longer inherits the cwd branch; an explicit `-b/--branch` pins it.
- **Oversized snapshot bodies** fail with an explicit size-limit error instead of a generic decode error.
- **Local dashboard summaries** exclude `decision_category='dry_run'` rows, so observer-mode would-denies no longer count as real critical actions.

hasandemirkiran:
@greptileai

internal/githubpolicy/classifier.go
```diff
@@ -0,0 +1,684 @@
...
+	}
+	verb := ""
+	if len(positional) > 1 {
+		verb = positional[1]
+	}
+	resource := githubRepoFromGhArgs(args)
+	if resource == "" && area == "api" && len(positional) > 1 {
+		// A literal REST endpoint names the repository it targets; -R only
+		// fills {owner}/{repo} placeholders.
+		resource = githubRepoFromAPIEndpoint(positional[1])
```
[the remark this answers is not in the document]
hasandemirkiran: Fixed in 6cd61ae: the gh positional parser now treats -X/--method/--request, -H/--header, --input, and the -q/-t short forms as value flags, so `gh api -X DELETE repos/acme/admin/issues/1` extracts `acme/admin` from the endpoint instead of falling back to the cwd repo. Covered with tests for -X, --method=, -H, and --input forms.

hasandemirkiran:
@greptileai

internal/githubpolicy/classifier.go
```diff
@@ -0,0 +1,685 @@
...
+		return ProviderAction{Action: readWriteAction("github.pr", isReadVerb(verb)), Resource: resource}
+	case "issue":
+		return ProviderAction{Action: readWriteAction("github.issue", isReadVerb(verb)), Resource: resource}
+	case "repo":
+		return ProviderAction{Action: readWriteAction("github.repo", isReadVerb(verb)), Resource: resource}
+	case "workflow":
+		mutating := verb == "run" || verb == "enable" || verb == "disable"
+		return ProviderAction{Action: apiAction(mutating), Resource: resource}
+	}
+	return ProviderAction{Action: apiAction(commandLooksMutating(raw)), Resource: resource}
```
michiosw: This treats unsupported `gh` areas as reads unless the raw command has HTTP method or body flags. Commands like `gh release create ...` and `gh secret set ...` are writes, but they would be recorded as `github.api.read`. A policy that allows reads and denies writes would miss the denial. Please map these areas explicitly or fail closed to `github.api.write` for unknown `gh` subcommands that are not known reads.

internal/githubpolicy/classifier.go
```diff
@@ -0,0 +1,685 @@
...
+		if strings.HasPrefix(arg, "--repo=") {
+			return normalizeGithubRepo(strings.TrimPrefix(arg, "--repo="))
+		}
+	}
+	return ""
+}
+
+var ghValueOptions = map[string]bool{
+	"-R": true, "--repo": true, "--hostname": true, "--jq": true, "-q": true,
+	"--template": true, "-t": true, "--paginate": true, "--cache": true, "--preview": true,
```
michiosw: `--paginate` is a boolean `gh api` flag, but this table makes the parser skip the next argument as its value. For `gh api --paginate repos/acme/admin/pulls`, the endpoint gets dropped and policy falls back to the cwd repo. That evaluates the wrong repository. Please separate boolean flags from value-taking flags.

internal/githubpolicy/classifier.go
```diff
@@ -0,0 +1,685 @@
...
+// ClassifyProviderActions maps a tool invocation to canonical GitHub actions.
+// gitContext is invoked lazily — only when a git command needs remote or
+// branch resolution — because deriving it shells out to git.
+func ClassifyProviderActions(toolName string, toolInput map[string]any, gitContext func() GitContext) []ProviderAction {
+	if gitContext == nil {
+		gitContext = func() GitContext { return GitContext{} }
+	}
+	switch strings.ToLower(toolName) {
+	case "bash", "shell":
+		command, _ := stringField(toolInput, "command")
```
michiosw: This only reads `command` and `cmd`, but the rest of the guard path also accepts `script`. A shell hook with `script: "git push ..."` would run a GitHub action without producing a policy evaluation or dry-run row. Please reuse the existing command input shape and include `script`.

michiosw (commented):
I found a few issues in the GitHub policy observer path that should be fixed before merge. The main risk is that some real GitHub actions are skipped or evaluated against the wrong repo/action, so the observer data becomes misleading and enforce mode would inherit the same mistakes.

internal/githubpolicy/classifier.go
```diff
@@ -0,0 +1,685 @@
...
+	}
+	positionals := gitPositionals(args)
+	if len(positionals) == 0 || positionals[0] != verb {
+		return ""
+	}
+	if len(positionals) < 2 {
+		// push/fetch/pull without an explicit remote still hit the network —
+		// they default to origin, so the policy evaluation must too.
+		if verb == "push" || verb == "fetch" || verb == "pull" {
+			return gitContext().RemoteByName["origin"]
```
michiosw, line 657: This resolves `origin` from the hook cwd even when the command uses `git -C` to target another checkout. For `git -C ../other-repo push origin main`, the dry-run decision is recorded for the wrong repo and branch. Please make git context resolution aware of the target directory selected by `-C`.

internal/githubpolicy/classifier.go
```diff
@@ -0,0 +1,685 @@
...
+var githubHostRE = regexp.MustCompile(`(?i)(^|[\s"'(<])((https?://)?(api\.)?github\.com[/:]|(ssh://)?git@github\.com[:/])`)
+
+// ClassifyProviderActions maps a tool invocation to canonical GitHub actions.
+// gitContext is invoked lazily — only when a git command needs remote or
+// branch resolution — because deriving it shells out to git.
+func ClassifyProviderActions(toolName string, toolInput map[string]any, gitContext func() GitContext) []ProviderAction {
+	if gitContext == nil {
+		gitContext = func() GitContext { return GitContext{} }
+	}
+	switch strings.ToLower(toolName) {
```
michiosw, line 75: The package contract says GitHub CLI, MCP, and API activity is evaluated locally, but GitHub MCP tool names fall through here. For example, `mcp__github__get_pull_request` is already a normalized managed GitHub fixture, but this classifier emits no action for it. Please add MCP classification or narrow the package contract until MCP support exists.

internal/guard/store/sqlite/store.go
```diff
@@ -1485,6 +1494,7 @@ select actions.session_id,
```
michiosw, line 1484: This filters dry-run rows out of session summaries, but the `Events` query below still returns every `request.decided` row. The dashboard buckets returned events by `decision`, so an observe-mode would-deny can appear as a real denied event in the session log. Please apply the same dry-run filter to the normal event list, or expose those rows in a separate dry-run view.

hasandemirkiran:
* **#279** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/279?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/279?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

michiosw (approved):

michiosw:
### Merge activity

* **Jun 15, 10:01 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/279).
* **Jun 15, 10:01 AM UTC**: @michiosw merged this pull request with [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/279).

PR #282, opened 2026-06-12

michiosw:
## Where We Are

`kontext guard start` could write Claude hook settings before it proved the daemon could bind its HTTP port. If startup failed on the listen step, Claude kept a Guard hook that pointed at a daemon that never came up.

## Where We Want To Go

Only install Claude hooks after the daemon has passed the fail-fast startup checks that can still abort launch. A listen failure must leave `~/.claude/settings.json` unchanged.

## How do we get there

Bind the HTTP listener before installing Claude hooks, then serve that bound listener instead of calling `ListenAndServe` directly. Add a regression test that occupies the target port, runs `runDaemon`, and asserts Claude settings are not created. Validation: `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check`.

### DIFF
--- a/internal/guard/cli/cli.go
+++ b/internal/guard/cli/cli.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


michiosw:
* **#282** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/282?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/282?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #284, opened 2026-06-13
[AUTOMATION] fix(clawpatch): ignore stale dashboard event fetch errors

michiosw:
## Where We Are

The guard dashboard can show an error from session A after the user has already switched to session B. The stale request is ignored on success, but the stale error still reaches the banner.

## Where We Want To Go

Only the active session should be allowed to update the event error state. Switching sessions should not surface a false error from an older request.

## How do we get there

Apply the same `selectedRef.current !== id` guard to the `fetchEvents(id)` error path in `web/guard-dashboard/src/App.tsx` that already exists on the success path. Rebuild the embedded dashboard assets so the shipped UI matches the source change. Validation: `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, `./scripts/build-guard-dashboard-assets.sh`, and `git diff --check` passed. `go test ./...` is blocked by unrelated flaky timing failures in `internal/guard/judge` (`TestStartLlamaServerHealthCheckAndStop`, `TestStartLlamaServerEarlyExitDoesNotWaitForStopTimeout`).


michiosw:
* **#284** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/284?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/284?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #290, opened 2026-06-14

michiosw:
## Where We Are

Claude hook install and uninstall rewrote `~/.claude/settings.json` and the backup file with `0644`. That makes a private Claude settings file readable by other local users after a guard hook change.

## Where We Want To Go

Keep the existing Claude settings permissions when we rewrite the file. If the file is created fresh, default it to `0600`.

## How do we get there

Read the current file mode before writing `settings.json` or its backup, then reuse that mode instead of hard-coding `0644`. Added regression coverage for both install and uninstall to assert the rewritten settings file and the backup stay `0600`. Verified with `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check`.

### DIFF
--- a/internal/guard/cli/cli.go
+++ b/internal/guard/cli/cli.go
@@ -440,7 +440,8 @@ func readClaudeSettings() (string, map[string]any, error) {
 }
 
 func backupFile(path, label string) error {
-	if _, err := os.Stat(path); os.IsNotExist(err) {
+	mode, err := fileModeOrDefault(path, 0o600)
+	if os.IsNotExist(err) {
 		return nil
 	} else if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


michiosw:
* **#290** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/290?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/290?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #292, opened 2026-06-15
[AUTOMATION] fix(localruntime): harden guard socket dir

michiosw:
## Where We Are

The guard daemon trusted any existing parent directory for its Unix socket under /tmp. Another local user could pre-create that directory with unsafe permissions or redirect it through a symlink.

## Where We Want To Go

The guard daemon should only start when the socket directory is a real directory owned by the current user and locked down to mode 0700.

## How do we get there

We harden internal/localruntime.EnsureSocketDir to create the directory, reject symlinks and non-directories, verify the owner UID, and chmod safe same-user directories to 0700 before the daemon listens. We added focused socket directory tests for create, chmod, and symlink rejection. Verification: go test ./..., go vet ./..., npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile, npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck, git diff --check.

### DIFF
--- a/internal/localruntime/socket.go
+++ b/internal/localruntime/socket.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"syscall"
 )
 
 func DefaultSocketPath() string {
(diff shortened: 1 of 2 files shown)


michiosw:
* **#292** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/292?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #294, opened 2026-06-15
fix(managedstream): stop terminal ingest retry storms

michiosw:
## Where We Are

Managed observe retried hosted ingest 400s as if smaller batches might fix them. An organization mismatch or schema/config validation error is terminal until inputs change, so one broken install could replay rejected ledger data and create an ingest storm.

## Where We Want To Go

Terminal hosted 400/422 validation failures should stop hot-looping without dropping unsent ledger rows. Payload-size responses should still get smaller-batch retry. A credential/config change should clear the terminal state and let streaming resume.

## How do we get there

Classify hosted ingest failures before retrying. Keep smaller-batch retry for 413 and explicit size-like 400/422 responses, pause terminal hosted validation failures in persisted cooldown state without advancing the cursor, and clear that state when the effective upload config changes. Added regressions for org mismatch, generic 400, 422 validation, auth failures, malformed cooldown state, credential refresh, scalar size-like validation, and persisted state shape.

Verified with go test ./..., go vet ./..., go test -race ./internal/managedstream ./internal/managedobserve, codex-review, and autoreview.

### DIFF
--- a/internal/managedstream/stream.go
+++ b/internal/managedstream/stream.go
@@ -3,6 +3,8 @@ package managedstream
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -28,6 +30,7 @@ const (
 	DefaultBatchLimit = 100
 	MaxPayloadBytes   = 192 * 1024
 	DefaultInterval   = 10 * time.Second
+	DefaultCooldown   = 15 * time.Minute
 
 	maxPayloadSessions = 100
 	maxPayloadActions  = 1000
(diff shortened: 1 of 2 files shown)


michiosw:
* **#294** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/294?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/294?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

michiosw:
Closing this version. The persisted cooldown approach is too broad and Greptile found that unchanged terminal failures can still replay after cooldown expiry, some config changes do not clear the pause, and the daemon keeps a fixed config/credential snapshot. I am replacing it with a leaner implementation.

PR #295, opened 2026-06-15
fix(managedstream): avoid terminal ingest retry storms

michiosw:
## Where We Are

PR #294 tried to stop terminal hosted ingest failures with persisted cooldown state. Greptile found that unchanged terminal failures could still replay after cooldown expiry, some config changes did not clear the pause, and the daemon kept a fixed credential/config snapshot.

## Where We Want To Go

Terminal hosted 400/422 validation failures should stop the smaller-batch retry path without advancing the cursor. Real payload-size responses should still retry with a smaller batch. A running daemon should use refreshed managed config and install tokens on the next stream flush.

## How do we get there

Classify hosted 400/422 responses by body before shrinking batches. Generic validation failures now return after one hosted attempt and leave stream state untouched. The daemon now rebuilds managed stream upload options for each flush, so refreshed config and install tokens are used without restarting.

Verified with:

- go test ./internal/managedstream ./internal/managedobserve
- go test ./...
- go vet ./...
- go test -race ./internal/managedstream ./internal/managedobserve

### DIFF
--- a/internal/managedobserve/daemon.go
+++ b/internal/managedobserve/daemon.go
@@ -110,19 +110,7 @@ func RunDaemon(ctx context.Context, opts DaemonOptions) error {
 	defer stopStream()
 	streamErr := make(chan error, 1)
 	go func() {
-		streamErr <- managedstream.Run(streamCtx, managedstream.Options{
-			DBPath:            dbPath,
-			StatePath:         opts.StreamStatePath,
-			CloudURL:          loadedConfig.Config.CloudURL,
-			OrganizationID:    loadedConfig.Config.OrganizationID,
-			InstallationID:    installationState.InstallationID,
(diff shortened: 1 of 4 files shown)


michiosw:
* **#295** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/295?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/295?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

michiosw:
Closing this version. Replacing it with a smaller implementation that addresses the Greptile feedback directly.

PR #296, opened 2026-06-15
fix(managedobserve): refresh rotated install tokens

michiosw:
## Where We Are

PR #295 was closed because it fixed ledger token rotation but left GitHub policy refresh on the startup install token. It also kept shrinking batches for generic hosted 400/422 validation errors, which can retry the same invalid payload shape instead of stopping.

## Where We Want To Go

A running managed daemon should use a rotated install token for both policy refresh and ledger streaming. Hosted validation errors should fail once without advancing the cursor. Real 413 payload-size responses should still shrink batches, and a single rejected record should not be resent forever.

## How do we get there

Reload managed config and the install token before each policy refresh and stream flush. Limit hosted shrink retries to 413, and advance past a one-record 413 after recording the diagnostic error.

Verified with:

- go test ./internal/managedobserve ./internal/managedstream
- go test ./...
- go vet ./...
- go test -race ./internal/managedobserve ./internal/managedstream

### DIFF
--- a/internal/managedobserve/daemon.go
+++ b/internal/managedobserve/daemon.go
@@ -33,20 +33,13 @@ type DaemonOptions struct {
 }
 
 func RunDaemon(ctx context.Context, opts DaemonOptions) error {
-	loadedConfig, err := managedconfig.Load()
+	runtimeConfig, err := loadManagedRuntimeConfig(ctx)
 	if err != nil {
-		if errors.Is(err, managedconfig.ErrNotManaged) {
-			return err
-		}
-		return fmt.Errorf("load managed config: %w", err)
+		return err
 	}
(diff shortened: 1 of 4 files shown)


michiosw:
* **#296** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/296?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/296?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

michiosw:
Closing this version. The replacement will keep the token-refresh fix but avoid the single-record 413 cursor skip that Greptile flagged.

PR #297, opened 2026-06-15
fix(managedobserve): refresh install token per flush ENG-446

michiosw:
## Where We Are

ENG-446 started from a concrete retry storm: hosted ingest returned terminal `400` responses for an org mismatch, and the daemon treated them like size-related batch failures. With four records in a batch, the client could send the same bad payload shape at batch size 4, then 2, then 1. That wastes API capacity and floods logs while the server is correctly rejecting the request.

PR #296 tried to fix that and token rotation together, but it also advanced the cursor after a hosted one-record `413`. Greptile correctly flagged that as data loss risk because a hosted `413` can be generic or transient.

## Where We Want To Go

For this PR, terminal hosted validation errors should stop the current flush immediately. A `400` should not shrink the batch, should not advance the cursor, and should leave the data retryable after the operator fixes stale config or token/org mismatch.

A running managed daemon should also pick up rotated install tokens without restart for both GitHub policy refresh and ledger streaming. That prevents stale startup credentials from causing repeated hosted failures after token rotation.

This is the minimal safe slice for ENG-446. It does not add quarantine, long cooldown state, or doctor UX. Those are larger follow-ups if we want the full Linear acceptance criteria.

## How do we get there

Reload managed config and resolve the install token before each policy refresh and before each stream flush. Keep the existing local oversized-record cursor advance, because that path is based on client-side payload limits we can prove locally.

Change hosted retry behavior so only `413` reduces batch size. A hosted `400` or `422` now returns after one request. A hosted one-record `413` also returns without advancing the cursor, so Greptile's data-loss concern is addressed.

Tests cover the reviewer-critical cases:

- hosted `400` gets one request and no cursor advance
- hosted one-record `413` does not advance the cursor
- stream flushes pick up a refreshed install token
- GitHub policy refresh picks up a refreshed install token

Verified with:

- `go test ./internal/managedobserve ./internal/managedstream`
- `go test -race ./internal/managedobserve ./internal/managedstream`
- `go test ./...`
- `go vet ./...`
- `git diff --check`


michiosw:
* **#297** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/297?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/297?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

internal/managedstream/stream.go
```diff
@@ -282,9 +282,7 @@ func reducedBatchLimit(limit int) int {
 }
 
 func shouldRetryWithSmallerBatch(statusCode int) bool {
-	return statusCode == http.StatusBadRequest ||
-		statusCode == http.StatusRequestEntityTooLarge ||
-		statusCode == http.StatusUnprocessableEntity
+	return statusCode == http.StatusRequestEntityTooLarge
```
hasandemirkiran, line 348: Narrowing this to `413` is the right call for the `400` case in ENG-446. An org/installation mismatch is batch-wide, so shrinking only burns three identical requests before giving up.

My concern is `422`. Because the shrink re-queries from the same cursor with a smaller limit, it narrows toward the oldest unsent records rather than skipping a specific bad one. Today, if a single record inside an otherwise-valid batch is rejected, shrinking lets the older valid records drain before the stream eventually wedges on the bad one. With this change a per-record rejection stops the whole flush at full batch size, so valid older records are blocked behind the poison record until an operator intervenes. There's no data loss either way, as the cursor doesn't advance, but it is a head-of-line availability regression.

That only bites if the hosted ingest endpoint actually returns `422` per-record. I'm comfortable dropping `400`; `422` is conventionally semantic/per-record validation, so I'd like to confirm the backend contract first. If we aren't sure, I'd suggest keeping `422` in the shrink set until the quarantine follow-up lands.

hasandemirkiran (commented):
Narrowing `400` out of the shrink path is plainly correct, but I'd like us to confirm the backend's `422` behaviour before dropping it as well, since that's the status most likely to be returned per-record. Removing it turns a (wasteful) partial recovery into a head-of-line stall.

One smaller point while we're here: now that the daemon owns the ticker loops, `managedstream.Run` no longer has any callers and `githubpolicy.Run` is only reached from its own test. They also no longer match the daemon's behaviour, since they don't refresh the token. I'd prefer we either remove them or have both loops share a single implementation, so the fix can't silently regress in the copy that's no longer exercised.

internal/managedstream/stream.go
michiosw: yes that's right. Verified the backend contract: ledger validation fails return 400, bad/revoked install tokens return 401, and 413 for payload too large - -  no 422 on the ledger path today. So 413 stays the only shrink-retry status, and keeping it 413-only is correct.

PR #300, opened 2026-06-16
[AUTOMATION] fix(clawpatch): harden guard socket directory

michiosw:
## Where We Are

The guard daemon creates its local Unix socket under a predictable `/tmp/kontext-guard-<uid>` directory. If that directory already exists with loose permissions, the daemon keeps using it unchanged.

## Where We Want To Go

The guard daemon should only listen from a same-user, non-symlink socket directory with `0700` permissions.

## How do we get there

Make `internal/localruntime.EnsureSocketDir` match the managed-observe hardening path: reject symlinks and non-directories, require same-user ownership, and force `0700` before listening. Added focused tests for permissive directories and symlink parents. Verified with `go test ./...`, `go vet ./...`, `npm exec --yes --package pnpm@10.0.0 -- pnpm install --frozen-lockfile`, `npm exec --yes --package pnpm@10.0.0 -- pnpm --dir web/guard-dashboard typecheck`, and `git diff --check`.

### DIFF
--- a/internal/localruntime/socket.go
+++ b/internal/localruntime/socket.go
@@ -1,9 +1,11 @@
 package localruntime
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+	"syscall"
 )
 
 func DefaultSocketPath() string {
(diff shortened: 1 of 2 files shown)


michiosw:
* **#300** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/300?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/300?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #301, opened 2026-06-17
feat(managedstream): add daemon heartbeats

michiosw:
## Where We Are

Managed-observe only streamed ledger batches when hook activity existed. An idle but healthy daemon had no hosted signal, and failed idle uploads could retry every stream tick.

## Where We Want To Go

Idle managed daemons should check in through the existing ledger endpoint without changing the hosted payload contract. A normal ledger upload should also count as a heartbeat.

## How do we get there

This adds a 60-second heartbeat interval to managedstream, persists heartbeat attempt/success timestamps in stream state, and posts an empty ledger batch when no actions are pending. Failed heartbeats persist the attempt timestamp before upload so blocked egress does not retry every 10 seconds. The existing `organization_id` payload field stays unchanged. Verified with `go test ./...`, `go vet ./...`, and `git diff --check`.

### DIFF
--- a/internal/managedobserve/daemon_test.go
+++ b/internal/managedobserve/daemon_test.go
@@ -167,28 +167,35 @@ func TestDaemonStreamsLedgerBatches(t *testing.T) {
 		t.Fatalf("Process() error = %v", err)
 	}
 
-	select {
-	case body := <-requests:
-		if body.OrganizationID != "" {
-			t.Fatalf("organization_id sent = %q, want omitted (token binds the org)", body.OrganizationID)
-		}
-		if body.InstallationID != "ins_0123456789abcdefghijklmnopqrstuv" {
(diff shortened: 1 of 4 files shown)


michiosw:
* **#301** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/301?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/301?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

michiosw:
@greptileai

hasandemirkiran (dismissed):

michiosw:
Linear: ENG-453 https://linear.app/cobrowser/issue/ENG-453/dash0-make-daemon-and-api-logs-tell-the-product-story

PR #302, opened 2026-06-18
fix(setup): clarify daemon wait state

michiosw:
## Where We Are

During self-serve setup, the terminal can sit quietly while the CLI waits for the background agent. That looks stuck even though setup is still working.

The final line also points users toward starting Claude Code, but the onboarding loop expects them to return to the dashboard and run the hello command shown there.

## Where We Want To Go

Make every setup state visible and end with the dashboard as the next step.

Users should see that the CLI is waiting for the daemon, then get a clear return-to-dashboard instruction once setup is ready.

## How do we get there

The setup output is grouped into Workspace, Mac, and Next sections. The daemon probe now shows a small spinner in terminals and a plain waiting line in non-interactive output.

The organization-managed refusal now names the system config path and says why setup cannot continue.

Validated with:
- go test ./internal/setup
- go test ./cmd/kontext ./internal/managedconfig ./internal/managedobserve
- go test ./...

### DIFF
--- a/internal/setup/setup.go
+++ b/internal/setup/setup.go
@@ -101,6 +101,8 @@ func Run(ctx context.Context, opts Options) error {
 		return err
 	}
 
+	fmt.Fprintln(opts.Stdout, "Kontext setup")
+
 	cloudURL := strings.TrimSpace(opts.CloudURL)
 	if cloudURL == "" {
 		cloudURL = DefaultCloudURL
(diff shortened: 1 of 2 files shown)


michiosw:
* **#302** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/302?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/302?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #305, opened 2026-06-18
fix(setup): reload owned launchagent on reinstall

michiosw:
## Where We Are

Re-running `kontext setup` over an existing self-serve install can write the new workspace config, then fail while reloading launchd. The Mac is left pointing at the new token and config, but the old LaunchAgent may still be loaded.

## Where We Want To Go

A setup re-run should replace the self-serve LaunchAgent cleanly before bootstrapping it again. Installing a new workspace token over an older self-serve workspace should behave like a normal reinstall.

## How do we get there

Unload the exact LaunchAgent plist that setup owns before calling `launchctl bootstrap`, instead of unloading only by service label. Add regression coverage that setup uses the owned plist path before bootstrap. Verified with `go test ./internal/setup` and `go test ./...`.


michiosw:
* **#305** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/305?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/305?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #307, opened 2026-06-18
fix(githubpolicy): most-specific-wins eval + endpoint subject (ENG-450)

hasandemirkiran:
Part of [ENG-450](https://linear.app/cobrowser/issue/ENG-450/fix-github-access-policy-qa-gaps-from-observe-mode-testing) — fixing GitHub access-policy QA gaps found during observe-mode testing. Pairs with the cloud PR in `kontext-security/kontext`.

## 1. Most-specific-wins evaluation
The local engine used flat deny-overrides + layered-AND: a broad org `DENY` short-circuited to deny before a more-specific user/endpoint `ALLOW` could be weighed (so a "block the repo, allow my endpoint" carve-out never worked). `Evaluate` now ranks matching rules by:
1. pinned dimensions (action/resource/branch) — more beats fewer
2. specific subject (user/agent/endpoint) over org
3. deny over allow on an exact tie

No matching rule ⇒ default deny. Mirrors the cloud reference evaluator.

## 2. Endpoint subject
The managed endpoint has no resolved user/application identity, so user/agent-layer rules can never match locally. The daemon **does** always know its own installation id (`ins_…`), which is stable across pkg updates and uninstall. That id is now stamped as the endpoint-layer subject (`Request.EndpointID`), threaded `daemon → runtimehost → guard server → policy provider`. Endpoint-scoped rules are how device policy is matched on the managed path.

## Tests
- `go test ./...` green; gofmt/vet clean
- New/updated `evaluate_test.go` coverage including the carve-out scenario expressed with endpoint rules

## Note
The synced GitHub policy is still **observe-only** (mode hardcoded cloud-side), so these decisions are recorded as dry-runs and don't block yet — enforce is a separate, deliberate step.

### DIFF
--- a/internal/githubpolicy/evaluate.go
+++ b/internal/githubpolicy/evaluate.go
@@ -2,7 +2,6 @@ package githubpolicy
 
 import (
 	"fmt"
-	"strings"
 )
 
 // Reason codes shared with the cloud evaluator.
@@ -15,13 +14,16 @@ const (
 // ApplicationID are the Kontext user and application subjects; the managed
 // endpoint's trusted identity is the service account + installation, so both
 // are typically empty (unresolved) and user/agent-layer rules then never
-// match their subject.
+// match their subject. EndpointID is the installation ("ins_…") identity of
+// this managed endpoint — it is always known locally, so endpoint-layer rules
+// are the way to scope policy to a specific device on the managed path.
 type Request struct {
 	Action        string
 	Resource      string
 	BranchOrRef   string
 	UserID        string
 	ApplicationID string
+	EndpointID    string
 }
 
 // MatchedRule records one rule that matched the request, and whether it was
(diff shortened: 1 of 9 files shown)


internal/githubpolicy/evaluate.go
```diff
@@ -166,13 +143,49 @@ func ruleMatchesRequest(rule *Rule, request Request) bool {
...
+// rulePrecedes reports whether candidate should win over the current best under
+// most-specific-wins: more pinned dimensions, then user/agent over org, then
+// deny over allow on an exact tie. Returning false on a full tie keeps the
+// earlier rule, so evaluation is deterministic in snapshot order.
+func rulePrecedes(candidate, best *Rule) bool {
+	if dc, db := ruleDimensions(candidate), ruleDimensions(best); dc != db {
+		return dc > db
 	}
-	return false
+	if lc, lb := layerRank(candidate.Layer), layerRank(best.Layer); lc != lb {
```
michiosw, line 154: this lets a same-specifity endpoint/user/agent allow beat an org deny. the snapshot contract still says any matching deny wins and org is the ceiling, so enforce mode can now allow an action the cloud policy says should be denied. can we keep the old deny/org veto semantics unless the cloud contract changes in the same rollout?

hasandemirkiran: Confirmed direction: we're moving to **most-specific-wins**, not keeping deny/org-veto. The key point is that the contract changes in the **same rollout**, so the CLI and cloud don't diverge:

- Paired cloud PR **kontext-security/kontext#588** updates the shared contract `packages/shared/src/github-policy-snapshot.ts` (the doc this package mirrors), the cloud evaluator `policy.service.ts` (`decideProviderAction`), and `docs/guides/github-access-control-catalog.md` — all to most-specific-wins.
- This PR's `evaluate.go` doc is already rewritten to most-specific-wins (more pinned dimensions → specific subject over org → deny on an exact tie; default-deny when nothing matches), and the stale `snapshot.go` "NOT most-specific-wins / any deny wins" note was removed.

So when enforce mode lands, both engines share the same semantics — the CLI won't allow anything the cloud would deny. The two PRs ship together (both on ENG-450); merging one without the other is the only way they'd drift, which we'll avoid. The "falls through to deny-wins" wording left in `evaluate.go` is just the tie-break within most-specific-wins (equal specificity → deny), not the old global deny-override.

michiosw: i’m fine with most-specific-wins as the product behavior, but then this PR also needs the cloud snapshot contract/producer updated. right now the shared contract still says any matching deny wins and org is the ceiling, while the CLI now lets same-resource endpoint/user/agent allow beat org deny. that mismatch is the issue.

michiosw (approved):

michiosw:
### Merge activity

* **Jun 19, 12:17 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/307).
* **Jun 19, 12:18 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/307) rebased this pull request as part of a merge.
* **Jun 19, 12:20 PM UTC**: @michiosw merged this pull request with [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/307).

PR #319, opened 2026-06-19
feat: add codex agent adapter

tumberger:
## Summary
- Register Codex as a first-class agent adapter.
- Add Codex hook decoding/encoding for session, tool, prompt, and turn-stop events.

## Review notes
- `UserPromptSubmit` is intentionally treated as a blocking event in the shared hook domain. In enforce mode, runtime/socket failures fail closed for `UserPromptSubmit`, matching Codex's ability to block prompts. Claude Guard user hooks still only install `PreToolUse`/`PostToolUse`, so this does not add a Claude user-hook install path.
- Codex `Stop` is treated as turn-scoped telemetry, not as a session-finalization event. Codex does not emit Claude's `SessionEnd`.

## Verification
- `go test ./...`

PR #318, opened 2026-06-19
refactor: introduce agent hook installer

tumberger:
## Summary
- Add shared hook merge/remove and JSON file helpers for agent integrations.
- Migrate Claude user hook settings onto the shared helpers without behavior changes.

## Verification
- `go test ./...`

PR #320, opened 2026-06-19
feat: install codex self-serve hooks

tumberger:
## Summary
- Install and remove Codex self-serve hooks in `~/.codex/hooks.json`.
- Install only Codex-supported event keys from this stack and use `Stop` as turn-scoped telemetry.
- Skip backup/write work when rerunning setup leaves the canonical Codex hooks file unchanged.

## Review notes
- Codex command hooks do not use `async: true`; current Codex behavior skips async handlers rather than running them in the background.
- Codex hooks are enabled by default in current Codex releases; `/hooks` trust review remains the user action setup needs to call out.

## Verification
- `go test ./...`

### DIFF
--- a/internal/codexmanaged/hooks.go
+++ b/internal/codexmanaged/hooks.go
@@ -1,0 +1,252 @@
+package codexmanaged
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/kontext-security/kontext-cli/internal/agenthooks"
+	"github.com/kontext-security/kontext-cli/internal/hook"
+)
+
+const (
+	DefaultKontextBinary = "/usr/local/bin/kontext"
+	DefaultHookTimeout   = 20
+)
+
+var SupportedEvents = []hook.EventAlias{
+	{Name: hook.HookSessionStart, Alias: "session-start"},
+	{Name: hook.HookPreToolUse, Alias: "pre-tool-use"},
+	{Name: hook.HookPostToolUse, Alias: "post-tool-use"},
+	{Name: hook.HookUserPromptSubmit, Alias: "user-prompt-submit"},
+	{Name: hook.HookStop, Alias: "stop"},
+}
+
+type Settings struct {
+	Hooks map[string][]MatcherGroup `json:"hooks"`
+}
+
+type MatcherGroup struct {
+	Matcher string    `json:"matcher"`
+	Hooks   []Handler `json:"hooks"`
+}
+
+type Handler struct {
+	Type    string   `json:"type"`
+	Command string   `json:"command"`
+	Args    []string `json:"args,omitempty"`
+	Timeout int      `json:"timeout,omitempty"`
+	Async   *bool    `json:"async,omitempty"`
+}
+
+// UserHooksPath returns ~/.codex/hooks.json, creating ~/.codex.
+func UserHooksPath() (string, error) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	codexDir := filepath.Join(home, ".codex")
+	if err := os.MkdirAll(codexDir, 0o755); err != nil {
+		return "", err
+	}
+	return filepath.Join(codexDir, "hooks.json"), nil
+}
+
+// UserHooksPathNoCreate returns ~/.codex/hooks.json without creating ~/.codex.
+func UserHooksPathNoCreate() (string, error) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(home, ".codex", "hooks.json"), nil
+}
+
+func ReadHooks(path string) (map[string]any, error) {
+	return agenthooks.ReadJSONFile(path, "Codex hooks")
+}
+
+func WriteHooks(path string, settings map[string]any) error {
+	return agenthooks.WriteJSONFile(path, settings)
+}
+
+func BackupHooks(path, label string) error {
+	return agenthooks.BackupFile(path, label)
+}
+
+func Template(kontextBinary string) Settings {
+	kontextBinary = strings.TrimSpace(kontextBinary)
+	if kontextBinary == "" {
+		kontextBinary = DefaultKontextBinary
+	}
+	settings := Settings{Hooks: make(map[string][]MatcherGroup, len(SupportedEvents))}
+	for _, event := range SupportedEvents {
+		settings.Hooks[event.Name.String()] = []MatcherGroup{{
+			Matcher: "",
+			Hooks: []Handler{{
+				Type:    "command",
+				Command: hookCommand(kontextBinary, event.Alias),
+				Timeout: DefaultHookTimeout,
+			}},
+		}}
+	}
+	return settings
+}
+
+func TemplateJSON(kontextBinary string) ([]byte, error) {
+	data, err := json.MarshalIndent(Template(kontextBinary), "", "  ")
+	if err != nil {
+		return nil, fmt.Errorf("marshal Codex hooks template: %w", err)
+	}
+	return append(data, '\n'), nil
+}
+
+func Validate(data []byte, kontextBinary string) error {
+	kontextBinary = strings.TrimSpace(kontextBinary)
+	if kontextBinary == "" {
+		kontextBinary = DefaultKontextBinary
+	}
+
+	var settings Settings
+	if err := json.Unmarshal(data, &settings); err != nil {
+		return fmt.Errorf("parse Codex hooks: %w", err)
+	}
+
+	var problems []string
+	if settings.Hooks == nil {
+		problems = append(problems, "hooks missing")
+	}
+	for _, event := range SupportedEvents {
+		if err := validateEvent(settings.Hooks[event.Name.String()], event, kontextBinary); err != nil {
+			problems = append(problems, err.Error())
+		}
+	}
+	if len(problems) > 0 {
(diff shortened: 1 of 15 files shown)


tumberger:
* **#321** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/321?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#320** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/320?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/320?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#319** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/319?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#318** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/318?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

tumberger:
@greptile check again

internal/codexmanaged/hooks.go
```diff
@@ -0,0 +1,253 @@
...
+	"github.com/kontext-security/kontext-cli/internal/agenthooks"
+	"github.com/kontext-security/kontext-cli/internal/hook"
+)
+
+const (
+	DefaultKontextBinary = "/usr/local/bin/kontext"
+	DefaultHookTimeout   = 20
+)
+
+var SupportedEvents = []hook.EventAlias{
```
hasandemirkiran, line 20: This is the main thing I want to talk through. We install six events, but two of them — `SessionEnd` and `PostToolUseFailure` — are not Codex events. I checked OpenAI's Codex hooks docs and the source-derived DeepWiki to be sure: Codex's events are SessionStart, UserPromptSubmit, PreToolUse, PostToolUse, Stop, PermissionRequest, PreCompact, PostCompact, and the Subagent pair. No SessionEnd, no PostToolUseFailure.

Two fallouts:
- The SessionEnd / PostToolUseFailure entries are dead weight — Codex never fires them, yet `Validate` requires them to be present. Codex seems to ignore unknown event keys when dispatching, but it is strict about unknown *output* fields (openai/codex#18887), so I would want to confirm it does not reject a hooks.json that carries keys it does not recognize.
- The one that worries me: this breaks session close for Codex. `localruntime.closeSessionEnd` only calls `CloseSession` on `SessionEnd`, and Codex never sends one, so a Codex session never gets finalized. Codex's nearest signal is `Stop`, but `Stop` fires per turn, not once per session, so it is not a straight rename.

So the real question: what is the intended Codex session lifecycle? My read is Codex wants its own event list rather than borrowing Claude's, with a deliberate call on how `Stop` maps onto our session-end. Happy to pair on it.

internal/codexmanaged/hooks.go
```diff
@@ -0,0 +1,253 @@
...
+			if handler.Type != "command" {
+				return fmt.Errorf("%s Kontext handler type = %q, want command", event.Name, handler.Type)
+			}
+			if len(handler.Args) > 0 {
+				return fmt.Errorf("%s Kontext handler args must be omitted", event.Name)
+			}
+			if handler.Timeout <= 0 {
+				return fmt.Errorf("%s Kontext handler timeout must be positive", event.Name)
+			}
+			if handler.Async != nil && *handler.Async {
```
hasandemirkiran, line 223: We never set `async` for Codex and this actively rejects it, so every Codex hook runs synchronously — including SessionStart, which is pure telemetry and cannot block anything. On Claude we mark SessionStart/SessionEnd async precisely so they do not add latency to startup, and Codex supports `async: true` on handlers too. Is forcing sync deliberate (does the ingest path need it?) or just inherited? If there is no reason for it, the non-blocking events probably want to be async so a slow daemon does not stall Codex coming up.

internal/setup/setup.go
```diff
@@ -683,6 +695,47 @@ func preflightLegacyUserHooks() error {
...
+	if err != nil {
+		return fmt.Errorf("check Codex hooks: %w\n\nFix or move ~/.codex/hooks.json, then rerun setup.", err)
+	}
+	if err := codexmanaged.MergeManagedHooks(settings, binary); err != nil {
+		return fmt.Errorf("check Codex hooks: %w\n\nFix or move ~/.codex/hooks.json, then rerun setup.", err)
+	}
+	return nil
+}
+
+func installCodexUserHooks(binary string) (string, error) {
```
hasandemirkiran, line 723: This backs up and rewrites hooks.json every run without checking whether anything actually changed. The merge is idempotent and the help text says re-running setup is safe, so in practice each `kontext setup` drops another timestamped backup and rewrites the file — run it a few times and ~/.codex fills up with backups. `cleanLegacyUserHooks` just above already guards exactly this with the `bytes.Equal(before, after)` early return; can we mirror that here?

internal/setup/setup.go
```diff
@@ -200,6 +205,13 @@ func Run(ctx context.Context, opts Options) error {
 	}
 	fmt.Fprintf(opts.Stdout, "  ✓ Claude Code managed hooks installed (%s)\n", settingsPath)
 
+	codexHooksPath, err := installCodexUserHooks(binary)
+	if err != nil {
+		return fmt.Errorf("install Codex hooks: %w\n\nFix or move ~/.codex/hooks.json, then rerun setup.", err)
+	}
+	fmt.Fprintf(opts.Stdout, "  ✓ Codex hooks installed (%s)\n", codexHooksPath)
+	fmt.Fprintln(opts.Stderr, "note: Codex hooks require review before they run; open `/hooks` in Codex to trust the Kontext hooks.")
```
hasandemirkiran, line 218: Good call pointing them at `/hooks` to trust the hooks. The gap: Codex hooks are still experimental and off by default — they only run with `[features] codex_hooks = true` in ~/.codex/config.toml. As written, we install them and tell the user to trust them, but if that flag is not set, nothing fires and we silently do not observe Codex at all. Could we detect the flag and warn, or at least name it in this note?

hasandemirkiran (commented):
Most of my comments land here; the headline is the Codex event set, flagged inline on SupportedEvents.

internal/codexmanaged/hooks.go
tumberger: Resolved in #320: `SupportedEvents` no longer installs `SessionEnd` or `PostToolUseFailure`; it installs Codex-supported events only, including `Stop`. The intended lifecycle is explicit now: `Stop` is turn-scoped telemetry and does not call `CloseSession`. Codex sessions are not finalized by pretending `Stop` is Claude `SessionEnd`; stale daemon-observed sessions remain handled by the existing cleanup path.

internal/codexmanaged/hooks.go
tumberger: Resolved by documenting the intentional behavior rather than changing code: current Codex command-hook behavior skips handlers marked `async: true`, so we should not emit async handlers for Codex. The #320 PR body now calls this out explicitly.

internal/setup/setup.go
tumberger: Resolved in #320: `installCodexUserHooks` now reads the existing file, merges the canonical hooks, marshals the result, and returns early when `bytes.Equal(before, after)` is true. I also added a setup idempotency assertion that rerunning setup creates zero Codex hook backups.

internal/setup/setup.go
tumberger: Resolved in #320 PR notes: current Codex hooks are enabled by default, and `codex_hooks` is now a deprecated alias for disabling/enabling hooks behavior. The setup note remains focused on the action users still need after install: open `/hooks` and trust the Kontext hooks.

tumberger:
@greptile review given all changes and updates and review

internal/setup/setup.go
hasandemirkiran: Thanks for digging in on this! One thing still nags me on the default — the OpenAI hook docs read as opt-in to me: `codex_hooks` is deprecated in favor of `[features].hooks`, but hooks seem to be off by default and silent until that key is set (https://developers.openai.com/codex/hooks). If that holds, someone who runs setup and trusts the hooks via `/hooks` but never sets `[features].hooks = true` would silently go unobserved. Are you seeing them on by default in the Codex version we target? If it's still opt-in, might be worth naming `[features].hooks` in the note so folks aren't left uncovered without realizing it.

hasandemirkiran (dismissed):
Did a functional test as well, works well

hasandemirkiran:
### Merge activity

* **Jun 29, 11:20 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/320).
* **Jun 29, 11:23 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/320) couldn't merge this pull request because a downstack PR #319 failed to merge.
* **Jun 29, 11:30 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/320).
* **Jun 29, 11:31 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/320) rebased this pull request as part of a merge.
* **Jun 29, 11:34 PM UTC**: @hasandemirkiran merged this pull request with [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/320).

hasandemirkiran (approved):

PR #336, opened 2026-06-23
feat: use managed hooks for self-serve cowork

michiosw:
## Where We Are

Self-serve setup still needed the same Claude managed-settings path as enterprise, while the old CoWork observer kept a second sandbox hook pipeline in the normal product path.

## Where We Want To Go

Self-serve and enterprise should share the managed-settings hook infra. Claude Code and CoWork should both report through that path, with CoWork labeled from managed hook context and no sandbox observer in normal installs.

## How do we get there

- Install self-serve Claude hooks through the managed-settings drop-in.
- Remove legacy user-level Kontext hooks before installing the drop-in, so setup cannot leave duplicate hook sources active.
- Refuse organization-managed Macs and unknown existing Claude drop-ins so self-serve cannot overwrite enterprise-owned hook files.
- Remove the sandbox CoWork observer from the daemon path; fallback code remains on feat/backup-cowork-sandbox-observer.
- Detect CoWork from anchored local-agent-mode session paths in the managed hook wrapper.
- Verified with `go test ./internal/setup` and `go test ./...`.


michiosw:
* **#336** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/336?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/336?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

michiosw:
Superseded by #337, which addresses all four P1 review findings (stale-hook refresh, install ordering before legacy cleanup, uninstall ownership check, deprecated `cowork_enabled` handling) on a clean single-commit branch.

PR #338, opened 2026-06-23
feat: use managed hooks for self-serve cowork

michiosw:
## Goal

Self-serve and enterprise share **one** managed-settings hook path. Claude Code and CoWork both report through it; CoWork is labeled from hook context; no sandbox observer in normal installs.

## Why removing the observer is safe (read first)

CoWork's Claude Code **honors the host managed-settings drop-in** — verified on a real machine: with the drop-in installed, both Claude Code and CoWork tool calls fire the managed hook. The old per-session sandbox observer existed for an earlier CoWork model where the VM only read session-local settings; that's no longer the case, so the observer is redundant. The managed hook fires inside CoWork and the wrapper labels it `cowork` from the `local-agent-mode-sessions` path.

(If a future CoWork change stops honoring the drop-in, the observer fallback is preserved on `feat/backup-cowork-sandbox-observer`.)

## Changes

- **Shared hook path** — `kontext setup` installs Claude hooks through the managed-settings drop-in (`…/managed-settings.d/20-kontext.json`), the same mechanism enterprise uses.
- **Content-based ownership** (`IsManagedSettingsDropIn`) — a drop-in is ours only when its **sole top-level key is `hooks`** and **every hook command is a Kontext managed-observe hook** (alias-matched, binary-path-agnostic). So:
  - a stale-binary-path drop-in from an older run is recognized and refreshed;
  - a file with foreign hooks **or extra top-level policy/metadata keys** is refused on setup and preserved on uninstall.
- **Safe ordering** — the authoritative drop-in is written **before** legacy `~/.claude/settings.json` hooks are removed, and that cleanup is non-fatal. A corrupt/unwritable legacy file can no longer leave a Mac with no hooks.
- **Observer removed** from the daemon path (fallback on `feat/backup-cowork-sandbox-observer`).
- **CoWork labeling** from anchored `local-agent-mode-sessions` paths in the managed hook wrapper. `cowork_enabled` in `managed.json` is accepted but ignored.

## Supersedes

Replaces #337 (and earlier #335/#336). All prior review findings are addressed and test-covered; this is a single clean commit.

## Validation

- `go test ./...`, `go vet ./...`, `gofmt` clean.


michiosw:
* **#338** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/338?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/338?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tumberger:
Small note on `cowork_enabled`: I think this is useful as a temporary migration guard, but we should not treat it as a long-term compatibility surface. In this PR the field no longer enables a separate CoWork observer; it only lets old configs fail loudly when the managed hook path is missing or disabled. That is a reasonable bridge for this rollout, but I would plan to supersede/remove the `LegacyCoworkEnabled` parse/runtime path after the managed-settings release has had time to migrate existing installs. The cleanup should keep the legacy user-hook removal and stale drop-in refresh behavior, since those are migration safety mechanisms rather than preserving the old observer model.

tumberger (approved):
lgtm with comment on cowork_enabled

PR #340, opened 2026-06-24
fix(setup): clean org-managed setup UX

michiosw:
## Where We Are

`kontext setup` gave a noisy failure when an organization-managed install was present. It printed usage, repeated the error, and left the user without a clean path to switch this Mac to self-serve.

## Where We Want To Go

Setup should make the choice obvious. If local organization-managed state exists, the user can remove it and continue, or decline and stop without writes. Non-interactive setup should fail once with clear cleanup steps.

## How do we get there

- Prompt interactive setup before writing self-serve state when `/Library/Application Support/Kontext/managed.json` exists.
- On yes, remove the local organization-managed config, install token, root Claude managed settings, stop the org-managed LaunchAgent, then continue setup.
- On no, stop cleanly and print the next action.
- Silence Cobra usage/errors for runtime setup failures and align uninstall output with setup status logs.
- Verified with `go test ./internal/setup ./cmd/kontext`, `git diff --check`, and `go test ./...`.

### DIFF
--- a/cmd/kontext/setup.go
+++ b/cmd/kontext/setup.go
@@ -10,8 +10,10 @@ func setupCmd() *cobra.Command {
 	var token, cloudURL string
 	var uninstall bool
 	cmd := &cobra.Command{
-		Use:   "setup",
-		Short: "Connect this Mac to your Kontext organization",
+		Use:           "setup",
+		Short:         "Connect this Mac to your Kontext organization",
+		SilenceUsage:  true,
+		SilenceErrors: true,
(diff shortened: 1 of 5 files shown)


michiosw:
* **#340** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/340?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/340?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #363, opened 2026-07-05
refactor: extract provider-neutral policy core into internal/providerpolicy

michiosw:
## Summary

- explain the user-facing change
- link the issue or context if relevant

## Verification

- [ ] `go test ./...`
- [ ] `go test -race ./...`
- [ ] `go vet ./...`
- [ ] `buf generate` if protobuf or generated code changed

## Release notes

- [ ] conventional commit title matches semver intent

PR #364, opened 2026-07-05
feat: hubspot provider policy — classifier, sync, dry-run + Cowork bash coverage

michiosw:
## Summary

- explain the user-facing change
- link the issue or context if relevant

## Verification

- [ ] `go test ./...`
- [ ] `go test -race ./...`
- [ ] `go vet ./...`
- [ ] `buf generate` if protobuf or generated code changed

## Release notes

- [ ] conventional commit title matches semver intent

### DIFF
--- a/internal/guard/app/server/policy.go
+++ b/internal/guard/app/server/policy.go
@@ -9,7 +9,61 @@ import (
 	"github.com/kontext-security/kontext-cli/internal/guard/judge"
 	guardpolicy "github.com/kontext-security/kontext-cli/internal/guard/policy"
 	"github.com/kontext-security/kontext-cli/internal/guard/risk"
+	"github.com/kontext-security/kontext-cli/internal/hubspotpolicy"
+	"github.com/kontext-security/kontext-cli/internal/providerpolicy"
 )
+
+// ProviderPolicyBinding wires one provider's synced policy into the decision
+// path: a snapshot source plus a classifier turning hook events into that
+// provider's canonical policy requests (subject fields are filled in by the
+// evaluator).
+type ProviderPolicyBinding struct {
+	Provider  string
+	Snapshots providerpolicy.SnapshotProvider
+	Classify  func(event risk.HookEvent) []providerpolicy.Request
+}
+
+// GithubPolicyBinding binds the GitHub classifier to a snapshot source.
+func GithubPolicyBinding(snapshots providerpolicy.SnapshotProvider) ProviderPolicyBinding {
+	return ProviderPolicyBinding{
+		Provider:  "github",
+		Snapshots: snapshots,
+		Classify: func(event risk.HookEvent) []providerpolicy.Request {
+			actions := githubpolicy.ClassifyProviderActionsWithCWD(event.ToolName, event.ToolInput, event.CWD, func(cwd string) githubpolicy.GitContext {
+				return githubpolicy.GitContextFromCWD(cwd)
+			})
+			requests := make([]providerpolicy.Request, 0, len(actions))
+			for _, action := range actions {
+				requests = append(requests, providerpolicy.Request{
+					Action:      action.Action,
+					Resource:    action.Resource,
+					BranchOrRef: action.BranchOrRef,
+				})
+			}
+			return requests
+		},
+	}
+}
+
+// HubspotPolicyBinding binds the HubSpot classifier to a snapshot source. The
+// Cowork connector registry is resolved lazily per event from the hook cwd.
+func HubspotPolicyBinding(snapshots providerpolicy.SnapshotProvider) ProviderPolicyBinding {
+	return ProviderPolicyBinding{
+		Provider:  "hubspot",
+		Snapshots: snapshots,
+		Classify: func(event risk.HookEvent) []providerpolicy.Request {
+			actions := hubspotpolicy.ClassifyProviderActions(event.ToolName, event.ToolInput, hubspotpolicy.ConnectorResolverForCWD(event.CWD))
+			requests := make([]providerpolicy.Request, 0, len(actions))
+			for _, action := range actions {
+				requests = append(requests, providerpolicy.Request{
+					Action:   action.Action,
+					Resource: action.Resource,
+				})
+			}
+			return requests
+		},
+	}
+}
 
 type PolicyProvider interface {
 	DecideHook(context.Context, risk.HookEvent) (risk.RiskDecision, error)
@@ -23,9 +77,10 @@ type RiskPolicyProvider struct {
 	judge        judge.Judge
 	policyEngine guardpolicy.Engine
 	policyConfig PolicyConfigProvider
-	githubPolicy githubpolicy.SnapshotProvider
+	// providerPolicies are the synced-policy bindings evaluated per event.
+	providerPolicies []ProviderPolicyBinding
 	// endpointID is this managed endpoint's installation ("ins_…") id, used as
-	// the endpoint-layer subject when evaluating synced GitHub policy.
+	// the endpoint-layer subject when evaluating synced provider policy.
 	endpointID string
 }
 
@@ -34,7 +89,7 @@ type RiskPolicyProviderOptions struct {
 	PolicyEngine         guardpolicy.Engine
 	PolicyConfig         guardpolicy.Config
 	PolicyConfigProvider PolicyConfigProvider
-	GithubPolicy         githubpolicy.SnapshotProvider
+	ProviderPolicies     []ProviderPolicyBinding
 	EndpointID           string
 }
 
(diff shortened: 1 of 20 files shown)


michiosw:
* **#364** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/364?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/364?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#363** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/363?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

internal/hubspotpolicy/classifier.go
```diff
@@ -0,0 +1,261 @@
...
+	"get_user_details":              {action: ActionAPIRead},
+	"get_organization_details":      {action: ActionAPIRead},
+	"search_owners":                 {action: ActionAPIRead},
+	"get_campaign_analytics":        {action: ActionAPIRead},
+	"get_campaign_asset_metrics":    {action: ActionAPIRead},
+	"get_campaign_contacts_by_type": {action: ActionAPIRead},
+	"get_content_analytics_report":  {action: ActionAPIRead},
+	"render_landing_page_ui":        {action: ActionAPIRead},
+	"tool_guidance":                 {action: ActionAPIRead},
+	"submit_feedback":               {action: ActionAPIRead},
```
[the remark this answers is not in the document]
michiosw: Fixed in df30da0 — `submit_feedback` now classifies as `hubspot.api.write` since it sends data to HubSpot (egress side effect), so a deny on `hubspot.api.write` will catch it.

internal/guard/app/server/policy.go
```diff
@@ -77,83 +132,88 @@ func (p RiskPolicyProvider) DecideHook(ctx context.Context, event risk.HookEvent
...
+				evaluations = append(evaluations, evaluation)
+			}
+		}
+		if len(evaluations) == 0 {
+			continue
+		}
+		all = append(all, risk.ProviderPolicyEvaluations{Provider: binding.Provider, Evaluations: evaluations})
+		if snapshot.Enforce() && enforcing == nil {
+			enforcing = &all[len(all)-1]
 		}
```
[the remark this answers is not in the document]
michiosw: Fixed in df30da0 — now points at a local copy (`g := group`) instead of `&all[len(all)-1]`, so a later append() reallocation can't leave the pointer at a stale backing array. (Not a live bug in observe mode — appended elements are copied and never mutated — but the footgun is gone.)

internal/guard/store/sqlite/dryrun_test.go
```diff
@@ -43,7 +43,7 @@ func TestSaveDecisionRecordsGithubDryRunRows(t *testing.T) {
 			Mode:           githubpolicy.ModeObserve,
 			Epoch:          5,
 			Hash:           "hash-5",
-		}},
+		}}}},
```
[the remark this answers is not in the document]
michiosw: Not a compilation failure — `githubpolicy.Evaluation`, `.Request`, `.EffectDeny` etc. are type *aliases* for the `providerpolicy` equivalents (internal/githubpolicy/policy.go: `type Evaluation = providerpolicy.Evaluation`). An alias is the identical type, so `[]githubpolicy.Evaluation` is assignable to the `[]providerpolicy.Evaluation` field — the `verify` CI job (go build + full test suite) passes, which proves it compiles. The githubpolicy alias reads naturally for a github-provider test row, so leaving as-is.

internal/guard/store/sqlite/dryrun_test.go
```diff
@@ -236,3 +236,77 @@ func TestSaveDecisionWithoutGithubPolicyAddsNoExtraRows(t *testing.T) {
...
+			Request:        githubpolicy.Request{Action: "hubspot.object.write", Resource: "deals"},
+			Result:         "deny",
+			ReasonCode:     githubpolicy.ReasonCodeDeny,
+			Reason:         "would deny",
+			DecidingRuleID: "rule-deny-deals",
+			Mode:           githubpolicy.ModeObserve,
+			Epoch:          2,
+			Hash:           "hash-2",
+			SchemaVersion:  "hubspot-policy-snapshot-v1",
+		}}}},
```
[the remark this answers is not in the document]
michiosw: These are type aliases (`internal/githubpolicy/policy.go`: `type Evaluation = providerpolicy.Evaluation`), so it compiled fine — but agreed the canonical names read better in a provider-neutral test. Switched the file to `providerpolicy.` in the latest push.

internal/guard/app/server/policy.go
```diff
@@ -77,83 +132,93 @@ func (p RiskPolicyProvider) DecideHook(ctx context.Context, event risk.HookEvent
...
+			if evaluation, ok := providerpolicy.Evaluate(snapshot, status, request); ok {
+				evaluations = append(evaluations, evaluation)
+			}
+		}
+		if len(evaluations) == 0 {
+			continue
+		}
+		group := risk.ProviderPolicyEvaluations{Provider: binding.Provider, Evaluations: evaluations}
+		all = append(all, group)
+		if snapshot.Enforce() && enforcing == nil {
```
hasandemirkiran: When two enforcing providers both produce evaluations for one event, only the first binding's group reaches `providerPolicyDecision` — a would-deny from the second provider is ignored if the first allows. Today the classifiers are disjoint (shell/git vs MCP tool names) and everything is observe-mode, so this can't fire, but the enforce branch is live code. "Any enforcing provider's deny wins" would be the safer invariant — cheap to do now (collect all enforcing groups, deny if any denies), or at least a TODO before enforce mode ships.

internal/managedobserve/daemon.go
```diff
@@ -273,39 +277,63 @@ func loadManagedConfig(ctx context.Context) (managedconfig.LoadedConfig, string,
...
 		cache.MarkFetchFailed(err)
-		opts.Diagnostic.Printf("github policy config reload: %v\n", err)
+		opts.Diagnostic.Printf("%s policy config reload: %v\n", config.ProviderKey, err)
+		return
+	}
+	snapshot, err := providerpolicy.FetchSnapshot(ctx, opts.PolicyHTTPClient, loadedConfig.Config.CloudURL, installToken, installationID, config)
+	if errors.Is(err, providerpolicy.ErrNotConfigured) {
+		// Contract: the cloud signals "provider deactivated / no rules" with an
+		// EMPTY snapshot (epoch 0, zero rules), never a 404. A 404 therefore
+		// means the route does not exist yet (older cloud during rollout —
```
hasandemirkiran, line 351: This comment frames a 404 as "route not deployed yet (older cloud) or a transient infra blip", but the test comment in daemon_test.go frames it as "org without an activated HubSpot provider". Behavior is identical either way (keep cache, mark stale, stay quiet), but the two framings disagree about the cloud contract — worth aligning so a future reader knows whether a permanent 404 is expected in steady state.

internal/managedobserve/daemon.go
```diff
@@ -142,7 +145,8 @@ func RunDaemon(ctx context.Context, opts DaemonOptions) error {
 
 	policyCtx, stopPolicyRefresh := context.WithCancel(ctx)
 	defer stopPolicyRefresh()
-	go runGithubPolicy(policyCtx, opts, policyCache, installationState.InstallationID)
+	go runProviderPolicy(policyCtx, opts, githubpolicy.Config, githubCache, installationState.InstallationID)
+	go runProviderPolicy(policyCtx, opts, hubspotpolicy.Config, hubspotCache, installationState.InstallationID)
```
hasandemirkiran: Every endpoint now polls the hubspot snapshot every 60s even for orgs that never activate HubSpot — a 404 per endpoint per minute, forever, and fleet-wide policy polling doubles with each provider added. Fine at this volume, but a longer retry interval after `ErrNotConfigured` would be a cheap follow-up as more providers stack up.

hasandemirkiran (commented):
Most of what it flagged was minor noise, but a few findings looked valid enough to pass along - left them inline. Nothing blocking from my side; the enforce-path one is the only one I'd consider addressing before enforce mode ships.

internal/guard/app/server/policy.go
michiosw: Sharp catch — fixed rather than TODO'd. `evaluateProviderPolicies` now returns every enforcing group and `providerPolicyDecision` denies if ANY enforcing evaluation denies (allow anchors on the first group). Added a two-enforcing-providers test proving a second provider's deny wins over the first's allow.

internal/managedobserve/daemon.go
michiosw: Aligned — the test comment now states the same contract as the daemon: deactivation is signaled by an EMPTY snapshot, never 404; a 404 means the route is not deployed (or a blip), so keep cache, mark stale, back off. A permanent 404 is expected steady state only for clouds that predate the provider's route.

internal/managedobserve/daemon.go
michiosw: Cheap enough to do now: while the cloud answers 404 the refresh loop backs off to every 10th tick (~10 min at the default 60s), resetting on the first non-404 answer. One-time ~10 min activation lag when a provider route first ships seemed acceptable.

PR #373, opened 2026-07-09
fix: managed-observe stale-daemon recovery — binary watchdog, status breadcrumb, doctor + --fix (ENG-516)

michiosw:
Ships the complete [ENG-516](https://linear.app/cobrowser/issue/ENG-516/brew-upgrade-leaves-managed-observe-daemon-on-old-binary-trace-export) fix. Originally a 3-PR Graphite stack (#373 → #374 → #375); the stacked children auto-merged into this branch (auto-merge arms on any non-draft PR and feature-branch bases have no review protection), so this PR now carries the full feature. Content is unchanged from the reviewed stack.

## Problem

`brew upgrade` swaps the binary but never restarts the launchd service — `KeepAlive` kept a deleted 0.13.0 binary running for ~38h with trace export silently stalled, while `kontext doctor` said `daemon: running`.

## What ships

1. **Binary watchdog** — daemon stats its own resolved executable every 2m and exits cleanly (→ KeepAlive respawns the new binary) on: file deleted (brew removes old Cellar dir), inode changed (in-place overwrite), or launch-path symlink retargeted while the old keg survives (`HOMEBREW_NO_INSTALL_CLEANUP`). Plus unconditional startup/error logging so the plist log is no longer 0 bytes.
2. **`daemon-status.json` breadcrumb** — version/pid/started_at, written once at startup, PID-alive freshness contract.
3. **Doctor enrichment + `--fix`** — `daemon: running (vX, pid N)` with stale-version WARNING (incl. unknown-version daemons that predate the breadcrumb — the first-upgrade case); heartbeat age (warn >5m); export cursor lag vs newest ledger event via read-only `sqlite.LedgerLag` (never migrates under a running daemon; warn >10m; never claims "up to date" with rows pending). `kontext doctor --fix` runs `launchctl kickstart -k` and only reports success after verifying the restarted daemon serves on the installed version.

## Verification

- `go test -race` green across `managedobserve`, `guard/store/sqlite`, `cmd`.
- 3-round adversarial review loop (Codex gpt-5.5, xhigh): rounds 1–2 findings all fixed with tests; round 3 verdict **MERGE_READY, zero findings**.
- New doctor smoke-tested against a live daemon.
- Post-release: live brew-upgrade respawn check rides the first release containing this.

### DIFF
--- a/cmd/kontext/main.go
+++ b/cmd/kontext/main.go
@@ -64,15 +64,42 @@ func main() {
 }
 
 func doctorCmd() *cobra.Command {
-	return &cobra.Command{
+	var fix bool
+	cmd := &cobra.Command{
 		Use:   "doctor",
 		Short: "Inspect local Kontext CLI setup",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			guardcli.PrintHookStatus(cmd.OutOrStdout())
-			managedobserve.PrintStatus(cmd.OutOrStdout())
+			staleDaemon := managedobserve.PrintStatus(cmd.OutOrStdout(), version)
+			if fix {
+				if !staleDaemon {
+					fmt.Fprintln(cmd.OutOrStdout(), "nothing to fix")
+					return nil
+				}
+				ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+				defer cancel()
+				if err := managedobserve.KickstartLaunchdKill(ctx, managedobserve.DefaultLabel()); err != nil {
+					fmt.Fprintf(cmd.ErrOrStderr(), "failed to restart managed-observe daemon: %v\n", err)
+					return err
+				}
+				// launchd accepting the kickstart is not a comeback: the new
+				// daemon can exit immediately (unreadable token, codesigning
+				// kill). Only report success once the restarted daemon is
+				// serving on the installed version.
+				waitCtx, waitCancel := context.WithTimeout(context.Background(), 10*time.Second)
+				defer waitCancel()
+				status, err := managedobserve.WaitForDaemonRestart(waitCtx, managedobserve.DefaultDBPath(), managedobserve.DefaultSocketPath(), version)
+				if err != nil {
+					fmt.Fprintln(cmd.ErrOrStderr(), "daemon has not come back within 10s — it may still be restarting; re-run `kontext doctor` in a minute and check ~/Library/Logs/Kontext/managed-observe.log")
+					return err
+				}
+				fmt.Fprintf(cmd.OutOrStdout(), "restarted managed-observe daemon (v%s, pid %d)\n", status.Version, status.PID)
+			}
 			return nil
 		},
 	}
+	cmd.Flags().BoolVar(&fix, "fix", false, "restart the managed-observe daemon when it is running a stale binary")
+	return cmd
 }
 
 func guardCmd() *cobra.Command {
--- a/cmd/kontext/main.go
+++ b/cmd/kontext/main.go
@@ -404,10 +404,11 @@ func managedObserveDaemonCmd() *cobra.Command {
 				timeout = parsed
 			}
 			return managedobserve.RunDaemon(context.Background(), managedobserve.DaemonOptions{
-				SocketPath:  socketPath,
-				DBPath:      dbPath,
-				IdleTimeout: timeout,
-				Diagnostic:  diagnostic.New(cmd.ErrOrStderr(), diagnostic.EnabledFromEnv()),
+				SocketPath:    socketPath,
+				DBPath:        dbPath,
+				IdleTimeout:   timeout,
+				Diagnostic:    diagnostic.New(cmd.ErrOrStderr(), diagnostic.EnabledFromEnv()),
+				BinaryVersion: version,
 				// "cli-" prefix lets the dashboard distinguish self-serve brew
 				// installs (no MDM deployment-version marker) from packages.
 				FallbackDeploymentVersion: "cli-" + version,
--- a/cmd/kontext/main.go
+++ b/cmd/kontext/main.go
@@ -88,7 +88,12 @@ func doctorCmd() *cobra.Command {
 				// serving on the installed version.
 				waitCtx, waitCancel := context.WithTimeout(context.Background(), 10*time.Second)
 				defer waitCancel()
-				status, err := managedobserve.WaitForDaemonRestart(waitCtx, managedobserve.DefaultDBPath(), managedobserve.DefaultSocketPath(), version)
+				status, err := managedobserve.WaitForDaemonRestart(
+					waitCtx,
+					managedobserve.DefaultDBPath(),
+					managedobserve.DefaultSocketPath(),
+					version,
+				)
 				if err != nil {
 					fmt.Fprintln(cmd.ErrOrStderr(), "daemon has not come back within 10s — it may still be restarting; re-run `kontext doctor` in a minute and check ~/Library/Logs/Kontext/managed-observe.log")
 					return err
--- a/internal/guard/store/sqlite/lag.go
+++ b/internal/guard/store/sqlite/lag.go
@@ -1,0 +1,61 @@
+package sqlite
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"os"
+	"time"
+)
+
+// LedgerLag reports export backlog without opening the store read-write:
+// newest is the max updated_at cursor timestamp in authorization_actions
+// (nil when the table is empty), pending counts rows past the given cursor.
+// Read-only on purpose — doctor must never migrate a database out from
+// under a running (possibly older) daemon.
+func LedgerLag(ctx context.Context, dbPath string, cursor *LedgerCursor) (newest *time.Time, pending int, err error) {
+	if _, err := os.Stat(dbPath); err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, 0, nil
+		}
+		return nil, 0, err
+	}
+	db, err := sql.Open("sqlite", "file:"+dbPath+"?mode=ro")
+	if err != nil {
+		return nil, 0, err
+	}
+	defer db.Close()
+	db.SetMaxOpenConns(1)
+	if _, err := db.ExecContext(ctx, "pragma busy_timeout = 5000"); err != nil {
+		return nil, 0, err
+	}
+
+	var rawNewest sql.NullString
+	if err := db.QueryRowContext(ctx, "select max(updated_at_cursor_key) from authorization_actions").Scan(&rawNewest); err != nil {
+		return nil, 0, err
+	}
+	if rawNewest.Valid && rawNewest.String != "" {
+		parsed, err := parseLedgerTimestamp(rawNewest.String)
+		if err != nil {
+			return nil, 0, err
+		}
+		newest = &parsed
+	}
+
+	query := "select count(*) from authorization_actions"
+	args := []any{}
+	if cursor != nil {
+		updatedAfter := ledgerTimestampCursorKeyFromTime(cursor.UpdatedAt)
+		if cursor.ActionID != "" {
+			query += " where updated_at_cursor_key > ? or (updated_at_cursor_key = ? and id > ?)"
+			args = append(args, updatedAfter, updatedAfter, cursor.ActionID)
+		} else {
+			query += " where updated_at_cursor_key > ?"
+			args = append(args, updatedAfter)
+		}
+	}
+	if err := db.QueryRowContext(ctx, query, args...).Scan(&pending); err != nil {
+		return nil, 0, err
+	}
+	return newest, pending, nil
+}
--- a/internal/managedobserve/binarywatch.go
+++ b/internal/managedobserve/binarywatch.go
@@ -1,0 +1,114 @@
+package managedobserve
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/kontext-security/kontext-cli/internal/diagnostic"
+)
+
+const (
+	envBinaryWatchInterval     = "KONTEXT_BINARY_WATCH_INTERVAL"
+	defaultBinaryWatchInterval = 2 * time.Minute
+)
+
+// startBinaryWatchdog exits after launchd KeepAlive leaves an old deleted
+// binary running across brew upgrades; a clean exit lets launchd respawn the
+// new binary.
+func startBinaryWatchdog(ctx context.Context, log diagnostic.Logger) <-chan struct{} {
+	replaced := make(chan struct{}, 1)
+	resolved, initialInfo, ok := binaryWatchdogConfig(log)
+	if !ok {
+		close(replaced)
+		return replaced
+	}
+	// Capture the stat seam once so the goroutine never reads package state
+	// concurrently with tests swapping the seams.
+	stat := statPath
+	go runBinaryWatchdog(ctx, resolved, initialInfo, binaryWatchInterval(), stat, log, replaced)
+	return replaced
+}
+
+func binaryWatchdogConfig(log diagnostic.Logger) (string, os.FileInfo, bool) {
+	if runtimeGOOS != "darwin" {
+		return "", nil, false
+	}
+	exe, err := executablePath()
+	if err != nil {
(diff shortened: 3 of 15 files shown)


michiosw:
Both review comments addressed in the latest push: the watchdog's log-once flag now resets on a successful stat (recurring error bursts stay visible over a long daemon run), and the inode-replace test writes a temp file and renames it over the original instead of remove+rewrite (guaranteed new inode, matches real installer behavior).

tumberger (commented):
Reviewed with two independent passes (Go style/idiom + business-level sanity of the recovery design). Overall verdict: **merge-ready**. Verified against the code: all three brew-upgrade shapes are caught (delete / inode change / symlink retarget), clean exit does respawn under the unconditional `KeepAlive <true/>` + 30s throttle plist, the export cursor only advances after a successful POST so a watchdog exit can't lose data, and generic stat errors never fire the watchdog (no crash-loop risk). Inline comments below on the few spots worth fixing.

internal/managedobserve/doctor.go
```diff
@@ -104,3 +166,111 @@ func describeScope(scope managedconfig.Scope) string {
...
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-ticker.C:
+		}
+	}
+}
+
+func pidAlive(pid int) bool {
+	if pid <= 0 || runtime.GOOS == "windows" {
```
tumberger: `pidAlive` returns `true` for `pid <= 0`, so a corrupt/partial `daemon-status.json` with `"pid":0` gets trusted as a live daemon — doctor would print its (possibly stale) version instead of falling into the safer unknown-version branch. Zero/negative PIDs are never real processes on Unix; suggest returning `false` for `pid <= 0` and keeping `true` only for the Windows can't-check case.

internal/guard/store/sqlite/lag.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		return nil, 0, err
+	}
+	db, err := sql.Open("sqlite", "file:"+dbPath+"?mode=ro")
+	if err != nil {
+		return nil, 0, err
+	}
+	defer db.Close()
+	db.SetMaxOpenConns(1)
+	if _, err := db.ExecContext(ctx, "pragma busy_timeout = 5000"); err != nil {
```
tumberger, line 29: Worth noting: the main store never sets `journal_mode`, so this DB runs on the default rollback journal, not WAL. Under a sustained write burst the daemon's single write connection can hold an EXCLUSIVE lock past this 5s busy_timeout, and doctor would print `export: ERROR` at exactly the moment an operator is diagnosing such a burst. Diagnostic UX only (fails safe), but switching the store to WAL in a follow-up would let this read run concurrently.

internal/managedobserve/status.go
```diff
@@ -0,0 +1,46 @@
...
+	"os"
+	"path/filepath"
+	"time"
+)
+
+// DaemonStatus is the on-disk breadcrumb a daemon leaves after its socket is
+// serving. It is written once at daemon startup; readers must verify the PID is
+// alive before trusting it. A stale file from a dead daemon is expected and
+// harmless.
+// ponytail: PID-reuse false positive possible; add a heartbeat-refreshed updated_at if it ever bites.
```
tumberger: `// ponytail:` — this tag doesn't exist anywhere else in the repo; looks like a leftover session marker. Suggest `// TODO:` or plain prose.

internal/managedobserve/binarywatch_test.go
```diff
@@ -0,0 +1,171 @@
...
+	t.Setenv(envBinaryWatchInterval, "10ms")
+	runtimeGOOS = "darwin"
+	path := writeWatchdogBinary(t)
+	executablePath = func() (string, error) { return path, nil }
+	evalSymlinksPath = filepath.EvalSymlinks
+	ctx, cancel := context.WithCancel(context.Background())
+
+	replaced := startBinaryWatchdog(ctx, diagnostic.Logger{})
+	cancel()
+	time.Sleep(30 * time.Millisecond)
```
tumberger: nit: sleep-then-assert-no-fire can pass vacuously on a slow machine (the goroutine may simply not have ticked yet). Low real flake risk at 10ms intervals, but a `done` channel closed when `runBinaryWatchdog` returns would make this deterministic.

tumberger (commented):
Second style pass against the `golang-code-style` conventions (samber/cc-skills-golang). One earlier nit retracted: `args := []any{}` in lag.go is correct under the explicit-initialization rule (nil slices serialize to `null`), so that comment has been removed. New findings inline. Overall verdict unchanged: merge-ready; the one change worth making from this pass is the `runBinaryWatchdog` options struct.

internal/managedobserve/binarywatch.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	info, err := statPath(resolved)
+	if err != nil {
+		logAlways(log, "binary watchdog eligibility: stat %s: %v\n", resolved, err)
+		return "", "", nil, false
+	}
+	return exe, resolved, info, true
+}
+
+func runBinaryWatchdog(ctx context.Context, exe, resolved string, initialInfo os.FileInfo, interval time.Duration, stat func(string) (os.FileInfo, error), resolve func(string) (string, error), log diagnostic.Logger, replaced chan<- struct{}) {
```
tumberger: This signature is 9 parameters on a ~230-char line — past the point where wrapping helps; the fix is fewer parameters. `binaryWatchdogConfig()` above already gathers exactly the values this needs (`exe`, `resolved`, `initialInfo`, plus the captured `stat`/`resolve` seams), so promoting its returns into a small struct would collapse this to `(ctx, cfg, interval, log, replaced)` and match the `homebrewUpdaterConfigValue` pattern in updater.go. The call at line 38 would shrink accordingly.

internal/managedobserve/binarywatch.go
```diff
@@ -0,0 +1,129 @@
...
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	loggedStatErr := false
```
tumberger, line 89: nit: `var loggedStatErr bool` — reserve `:=` for non-zero values; `var` signals "starts at zero."

internal/managedobserve/doctor.go
```diff
@@ -53,13 +86,41 @@ func PrintStatus(out io.Writer) {
 		fmt.Fprintf(out, "  WARNING: install token is not readable (%v) — the agent cannot stream; re-run `kontext setup` or unlock your login keychain\n", err)
 	}
 
-	if conn, err := net.DialTimeout("unix", DefaultSocketPath(), 500*time.Millisecond); err == nil {
+	if conn, err := opts.Dial("unix", opts.SocketPath, 500*time.Millisecond); err == nil {
 		conn.Close()
-		fmt.Fprintln(out, "  daemon: running")
+		if status := LoadDaemonStatus(opts.DBPath); status != nil && pidAlive(status.PID) {
+			fmt.Fprintf(out, "  daemon: running (v%s, pid %d)\n", status.Version, status.PID)
+			if comparableVersion(status.Version) && comparableVersion(installedVersion) && status.Version != installedVersion {
```
tumberger, line 93: nit: three-operand condition hiding the actual business rule — named booleans read better, e.g.
```go
versionsKnown := comparableVersion(status.Version) && comparableVersion(installedVersion)
if versionsKnown && status.Version != installedVersion {
```

cmd/kontext/main.go
```diff
@@ -64,15 +64,42 @@ func main() {
...
+					fmt.Fprintf(cmd.ErrOrStderr(), "failed to restart managed-observe daemon: %v\n", err)
+					return err
+				}
+				// launchd accepting the kickstart is not a comeback: the new
+				// daemon can exit immediately (unreadable token, codesigning
+				// kill). Only report success once the restarted daemon is
+				// serving on the installed version.
+				waitCtx, waitCancel := context.WithTimeout(context.Background(), 10*time.Second)
+				defer waitCancel()
+				status, err := managedobserve.WaitForDaemonRestart(waitCtx, managedobserve.DefaultDBPath(), managedobserve.DefaultSocketPath(), version)
```
tumberger: nit: ~130-char call with 4 arguments — break one argument per line:
```go
status, err := managedobserve.WaitForDaemonRestart(
	waitCtx,
	managedobserve.DefaultDBPath(),
	managedobserve.DefaultSocketPath(),
	version,
)
```

internal/managedobserve/doctor_test.go
```diff
@@ -37,3 +43,316 @@ func TestPrintStatusReportsInstallationLoadError(t *testing.T) {
...
+		Now: func() time.Time { return e.now },
+	}
+}
+
+func (e doctorTestEnv) writeDaemonStatus(t *testing.T, pid int, version string) {
+	t.Helper()
+	if err := os.MkdirAll(filepath.Dir(DaemonStatusPath(e.dbPath)), 0o700); err != nil {
+		t.Fatal(err)
+	}
+	data := []byte(`{"version":"` + version + `","pid":` + itoa(pid) + `,"started_at":"2026-07-09T12:00:00Z"}` + "\n")
```
tumberger: nit: hand-concatenated JSON across five fragments is fragile (quoting/escaping lives in your head). `json.Marshal(DaemonStatus{Version: version, PID: pid, StartedAt: ...})` keeps the test in sync with the real struct — or `fmt.Sprintf` at minimum.

internal/managedobserve/doctor.go
```diff
@@ -53,13 +86,41 @@ func PrintStatus(out io.Writer) {
...
+			}
+		} else {
+			fmt.Fprintln(out, "  daemon: running")
+			// A serving daemon with no live status breadcrumb predates the
+			// breadcrumb feature — which makes it older than the installed
+			// binary by definition. This is exactly the first upgrade into
+			// this feature, so it must be fixable; a verified restart of an
+			// already-current daemon is the harmless worst case.
+			if comparableVersion(installedVersion) {
+				fmt.Fprintf(out, "  WARNING: daemon version is unknown — it likely predates v%s; run 'kontext doctor --fix' to restart it\n", installedVersion)
```
tumberger, line 105: Concern about the migration story this branch represents: the watchdog ships in the *new* binary, so it can't help the daemons that are stale today. The existing 24h self-updater covers part of the transition — daemons that haven't been manually upgraded will run `brew upgrade` themselves at their next tick, see a change, and exit for respawn. But any machine where someone runs `brew upgrade` manually *before* that tick hits the exact incident mechanism: the old daemon's updater then sees "already up to date" (`changed == false` in updater.go) and never exits — stale until `kontext doctor --fix`, reboot, or logout. Updater-ineligible installs (non-user scope, `KONTEXT_NO_UPDATE_CHECK`, no brew) are also stuck.

Is there a way to close that residual gap without the manual step? A few candidates:
- **Hook-side stale detection**: after an upgrade, hooks already execute the new binary while the daemon is old, and the hook path in `lifecycle.go` already kickstarts when the socket is down — could it also compare the daemon's breadcrumb/served version against its own and `kickstart -k` on mismatch (rate-limited)? That would self-heal on the first session after upgrade, no user action.
- **Brew formula `post_install`** doing the kickstart, so the same manual upgrade that creates the staleness also fixes it.
- Piggybacking on any interactive `kontext` command (e.g. `setup`, `status`) to opportunistically fix a detected stale daemon.

If the manual step stays for that path, worth a prominent release-notes callout. Happy with the code either way — this is a rollout question, not a correctness one. (Every upgrade *after* this release is fully handled by the watchdog within ~2 minutes.)

michiosw:
* **#373** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/373?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/373?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

internal/managedobserve/doctor.go
michiosw: Fixed in 5454dca. pidAlive now rejects non-positive PIDs before the Windows fallback, with regression coverage for 0 and -1.

internal/guard/store/sqlite/lag.go
michiosw: Agreed that this can surface export: ERROR under an unusually long write lock. I am leaving journal mode unchanged here: the diagnostic fails safe, and moving a fleet database to WAL is a separate storage change with compatibility and rollout testing. We can track it independently if contention shows up operationally.

internal/managedobserve/status.go
michiosw: Fixed in 5454dca. The repo-specific session marker is removed.

internal/managedobserve/binarywatch_test.go
michiosw: Fixed in 5454dca. The test now runs the watchdog directly, waits on a completion channel after cancellation, and separately verifies that no replacement signal was sent.

internal/managedobserve/binarywatch.go
michiosw: Fixed in 5454dca. The executable paths, initial file identity, interval, and captured seams now live in a private binaryWatchdogConfigValue; runBinaryWatchdog is down to context, config, logger, and result channel.

internal/managedobserve/binarywatch.go
michiosw: I am leaving this as := false. Both forms are idiomatic Go, and this repository uses explicit false initialization throughout; there is no repository convention reserving := for non-zero values. This would be style-only churn.

internal/managedobserve/doctor.go
michiosw: I am leaving the condition inline. The three predicates are short and directly express comparability plus mismatch; an extra local would restate the condition without clarifying behavior.

cmd/kontext/main.go
michiosw: Fixed in 5454dca. The restart verification call is now one argument per line.

internal/managedobserve/doctor_test.go
michiosw: Fixed in 5454dca. The fixture now writes the real DaemonStatus through writeJSONBreadcrumb, so escaping and schema tags stay aligned with production.

internal/managedobserve/doctor.go
michiosw: Confirmed. The watchdog cannot self-heal a daemon that has not restarted onto this release. I documented the one-time transition in the README: run kontext doctor --fix when doctor reports an old or unknown daemon version; subsequent upgrades recover automatically. I am not adding permanent hook-path restart and rate-limit logic or formula lifecycle behavior to this PR. If automatic first-adoption fleet recovery becomes an acceptance criterion, that should be a targeted rollout change.

internal/managedobserve/doctor_test.go
michiosw: Fixed in 5454dca. The fixture now writes the real DaemonStatus through writeJSONBreadcrumb, so escaping and schema tags stay aligned with production.

internal/managedobserve/doctor.go
michiosw: Confirmed. The watchdog cannot self-heal a daemon that has not restarted onto this release. I documented the one-time transition in the README: run kontext doctor --fix when doctor reports an old or unknown daemon version; subsequent upgrades recover automatically. I am not adding permanent hook-path restart and rate-limit logic or formula lifecycle behavior to this PR. If automatic first-adoption fleet recovery becomes an acceptance criterion, that should be a targeted rollout change.

michiosw:
Review follow-up is pushed in 5454dca. Beyond the inline fixes, this closes the watchdog initialization race where Homebrew could remove the running executable before the watchdog captured its baseline; initial resolve/stat ENOENT now triggers an immediate clean exit for launchd restart. Validation: go test ./... plus race tests for managedobserve, sqlite store, and cmd/kontext.

tumberger (approved):
lgtm

michiosw:
### Merge activity

* **Jul 13, 9:45 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/373).
* **Jul 13, 9:46 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/373) rebased this pull request as part of a merge.

PR #377, opened 2026-07-09
fix(guard): redact command summaries with the shared payloadcapture ruleset

tumberger:
## Why

Command and request summaries used a private credential matcher that diverged from the versioned redaction ruleset used for captured payloads. Moving summaries directly to `rules/1` exposed compatibility gaps, consumed shell separators, and ran the full redactor twice on command-bearing events.

## What

- Ship redaction ruleset `rules/2` and keep all redaction detection in `internal/payloadcapture/ruleset.json`.
- Add data-defined replacement templates so rules can redact values while preserving structural captures such as assignment keys, authorization schemes, shell separators, and query parameter delimiters.
- Export `payloadcapture.RedactText` and use it as the only summary redactor.
- Compute command redaction once and reuse the result for `CommandSummary` and `RequestSummary`.
- Fail closed to a content-free summary marker for commands larger than 64 KiB; deterministic classification continues to inspect the complete raw command.
- Truncate summaries at a valid UTF-8 boundary.

## Ruleset v2 coverage

- `password`, `pwd`, and `pass` assignments and flags
- access-key and access-token assignments
- plain `Authorization` values and short or long Bearer/Basic values
- credential-bearing query parameters without consuming `?` or `&`
- JSON credential properties embedded in flat command strings
- `-u` / `--user` `user:password` values
- known-prefix vendor credentials such as AWS, GitHub, OpenAI, Slack, npm, GCP, and JWTs

Generic entropy-only matching is intentionally excluded: treating every long opaque string as a secret would redact hashes, IDs, and other ordinary command arguments. Known credential prefixes and credential-bearing structures remain covered.

## Contract changes

- Captured payloads are stamped with `redactorVersion: rules/2`.
- Rule replacements use `$n` capture expansion, declared in the shared ruleset artifact.
- Exact redacted bytes and stored hashes can change where v2 preserves structure or adds coverage.
- Classification still runs on raw command text so redaction cannot blind credential or risk signals.

## Verification

- Shared coverage vectors pin every rule and exact structural output for v2 cases.
- Normalizer tests cover legacy regressions, JSON, curl userinfo, separator preservation, size limits, and UTF-8 truncation.
- Judge-input tests verify the preserved command structure.


internal/guard/risk/normalizer.go
```diff
@@ -307,15 +307,17 @@ func summarizeRequest(toolName, path, command string) string {
...
 func redact(value string) string {
-	value = credentialAssignment.ReplaceAllString(value, "$1=[redacted-credential]")
-	value = credentialHeader.ReplaceAllString(value, "${1}${2}[redacted-credential]")
-	value = credentialBearer.ReplaceAllString(value, "Bearer [redacted-credential]")
-	value = credentialShape.ReplaceAllString(value, "[redacted-credential]")
-	if len(value) > 240 {
-		return value[:240] + "..."
-	}
-	return value
+	redacted, _ := payloadcapture.RedactText(value)
```
michiosw, line 335: This loses credential coverage we already had. `ACCESS_KEY=<secret>`, plain `Authorization: <value>`, and short Bearer values now reach the summaries unredacted. Please preserve those cases and add regression tests before switching fully to the shared redactor.

internal/guard/risk/normalizer.go
```diff
@@ -307,15 +307,17 @@ func summarizeRequest(toolName, path, command string) string {
...
+	if len(redacted) > 240 {
```
michiosw: This replacement also consumes separators such as spaces and semicolons. For example, `printf ok;TOKEN=secret echo next` becomes roughly `printf ok[REDACTED_SECRET] echo next`, which changes the command structure shown to the judge. We need to preserve the separator while removing the credential.

internal/guard/risk/normalizer.go
```diff
@@ -307,15 +307,17 @@ func summarizeRequest(toolName, path, command string) string {
...
+		return redacted[:240] + "..."
```
michiosw: We now run all 14 redaction rules over the complete command twice, then keep only 240 bytes. On this branch, an 8 MB command takes about 6.3 seconds and allocates 587 MB. Please redact once and reuse the result, with a safe input-size limit for unusually large commands.

michiosw (changes requested):
The shared redactor is the right direction, but this currently introduces credential leaks, changes the command structure passed to the judge, and makes large commands too expensive for the hook path. Requesting changes until those regressions are addressed.

tumberger:
* **#377** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/377?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/377?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

internal/guard/risk/normalizer.go
tumberger: Restored in rules/2 (fe4c392): `sensitive-assignment` now covers `access[_-](key|token)` plus `pwd`/`pass` aliases, a new `authorization-header` rule handles plain `Authorization: <value>` with or without a scheme, and Bearer values redact at any non-empty length. Regression tests cover each shape at both layers: exact-output cases in `TestNormalizeRedactsCredentialValuesFromSummaries` and coverage vectors with `expected` assertions in `redaction-coverage-vectors.json`.

A follow-up (e53ceff) closes one more gap found while re-reviewing this: a value opening a quote that never closes (e.g. `TOKEN="unterminated-secret`) matched nothing under the first rules/2 grammar and leaked, where rules/1 redacted it. The value grammar now has a last-resort branch for that shape, with vectors and summary tests.

internal/guard/risk/normalizer.go
tumberger: Fixed at the ruleset level in rules/2 (fe4c392): rules capture the key/separator prefix and use `$1[REDACTED_SECRET]` replacement templates, so `printf ok;TOKEN=secret echo next` now becomes `printf ok;TOKEN=[REDACTED_SECRET] echo next` — separator, key name, and `=` all survive. Covered for `;`, `|`, outer quotes, and query-param delimiters in `risk_test.go`, and end-to-end at the judge-input layer in `TestJudgeInputPreservesRedactedCommandSeparators`.

internal/guard/risk/normalizer.go
tumberger: Fixed in 693ecb8: redaction runs once in `summarizeCommand` and the result is reused for both `CommandSummary` and `RequestSummary`. Inputs over 64 KiB fail closed to `[command omitted: exceeds summary limit]` without touching the regexes — omitted rather than truncated, because truncating before redaction could split a secret match and expose its prefix. Classification still runs on the full raw command (there's a test asserting an oversized command still classifies as destructive/production).

`BenchmarkNormalizeHookEventCommandSummary` covers 240 B–1 MiB: at the 64 KiB cap, normalization is ~21 ms / ~2.8 MB allocated (M4 Pro), and anything larger skips redaction entirely.

tumberger:
Server-side counterpart is up: kontext-security/kontext#655 syncs the canonical `redaction-ruleset` artifact (TS object, JSON mirror, and coverage vectors) to rules/2, so the cross-repo mirror contract on `ruleset.json` holds again. The JSON mirror and coverage vectors there are byte-for-byte copies of this PR's final state, and the shared zod schema now validates the `replacement` templates (must contain the placeholder), mirroring the Go-side compile check. Ingest already treats `redactorVersion` as opaque, so rules/2 envelopes need no server changes beyond the artifact sync.

michiosw (approved):
Re-reviewed the current head. The credential coverage, command structure, and performance regressions are fixed and covered by focused tests. Approved; keep the CLI release behind kontext#655 so the canonical rules/2 artifact lands first.

PR #382, opened 2026-07-14
feat(policy): add Go Cedar evaluator conformance

tumberger:
## Why

The endpoint needs a deterministic local Cedar boundary before policy retrieval or production wiring is introduced.

## What

- parses policy once and validates unique non-empty `@id` plus supported `@ask` metadata at load time
- implements the User / ToolUse / exact tool / complete JSON-input request contract
- defines binary64 JSON-number semantics and bounded JSON-to-Cedar conversion
- pins hashing, ASK behavior, and portable fixtures shared byte-for-byte with TypeScript
- verifies every supported hook adapter preserves the required input

## Verification

- full `go test -race ./...`
- `go vet ./...`
- byte-identical cross-runtime fixture comparison

## Change size

13 files, +1,689/-2 — evaluator, conversion, hashing, hook conformance, fixtures, and tests.

Tracks ENG-531.

### DIFF
--- a/internal/cedareval/context.go
+++ b/internal/cedareval/context.go
@@ -1,0 +1,251 @@
+package cedareval
+
+import (
+	"encoding/json"
+	"fmt"
+	"math"
+	"sort"
+	"strconv"
+	"strings"
+
+	cedar "github.com/cedar-policy/cedar-go"
+)
+
+const (
+	ContextMaxDepth  = 32
+	ContextMaxValues = 10_000
+	maxSafeInteger   = int64(9_007_199_254_740_991)
+)
+
+type ContextDiagnostic struct {
+	Code string `json:"code"`
+	Path string `json:"path"`
+}
+
+type ConversionError struct {
+	Code string
+	Path string
+	msg  string
+}
+
+func (e *ConversionError) Error() string {
+	return e.msg
+}
+
+type conversionState struct {
+	diagnostics []ContextDiagnostic
+	valuesSeen  int
+}
+
+func ConvertToolInput(toolInput map[string]any) (cedar.Record, []ContextDiagnostic, error) {
+	state := conversionState{diagnostics: make([]ContextDiagnostic, 0)}
+	value, err := convertValue(toolInput, "", 0, &state, false)
+	if err != nil {
+		return cedar.Record{}, nil, err
+	}
+	record, ok := value.(cedar.Record)
+	if !ok {
+		return cedar.Record{}, nil, newConversionError(
+			"invalid_json_value",
+			"",
+			"cedar context root must be an object",
+		)
+	}
+	sort.Slice(state.diagnostics, func(i, j int) bool {
+		return state.diagnostics[i].Path < state.diagnostics[j].Path
+	})
+	return record, state.diagnostics, nil
+}
+
+func convertValue(value any, path string, depth int, state *conversionState, insideSet bool) (cedar.Value, error) {
+	if depth > ContextMaxDepth {
+		return nil, newConversionError(
+			"maximum_depth_exceeded",
+			path,
+			fmt.Sprintf("cedar context exceeds depth %d", ContextMaxDepth),
+		)
+	}
+
+	state.valuesSeen++
+	if state.valuesSeen > ContextMaxValues {
+		return nil, newConversionError(
+			"maximum_values_exceeded",
+			path,
+			fmt.Sprintf("cedar context exceeds %d values", ContextMaxValues),
+		)
+	}
+
+	if value == nil {
+		if insideSet {
+			return nil, newConversionError(
+				"null_in_set",
+				path,
+				"json null cannot be represented in a cedar set",
+			)
+		}
+		state.diagnostics = append(state.diagnostics, ContextDiagnostic{
+			Code: "null_omitted",
+			Path: path,
+		})
+		return nil, nil
+	}
+
+	switch v := value.(type) {
+	case string:
+		return cedar.String(v), nil
+	case bool:
+		return cedar.Boolean(v), nil
+	case json.Number:
+		return convertNumber(v.String(), path)
+	case float64:
+		return convertFloat(v, path)
+	case float32:
+		return convertFloat(float64(v), path)
+	case int:
+		return convertSigned(int64(v), path)
+	case int8:
+		return convertSigned(int64(v), path)
+	case int16:
+		return convertSigned(int64(v), path)
+	case int32:
+		return convertSigned(int64(v), path)
+	case int64:
+		return convertSigned(v, path)
+	case uint:
+		return convertUnsigned(uint64(v), path)
+	case uint8:
+		return convertUnsigned(uint64(v), path)
+	case uint16:
+		return convertUnsigned(uint64(v), path)
+	case uint32:
+		return convertUnsigned(uint64(v), path)
+	case uint64:
+		return convertUnsigned(v, path)
+	case []any:
+		values := make([]cedar.Value, len(v))
+		for i, item := range v {
+			converted, err := convertValue(item, pointer(path, strconv.Itoa(i)), depth+1, state, true)
+			if err != nil {
+				return nil, err
+			}
+			if converted == nil {
+				return nil, newConversionError(
+					"null_in_set",
+					pointer(path, strconv.Itoa(i)),
+					"null cannot appear in a cedar set",
+				)
+			}
+			values[i] = converted
+		}
+		return cedar.NewSet(values...), nil
+	case map[string]any:
+		if _, ok := v["__entity"]; ok {
+			return nil, reservedEscapeError(path)
+		}
+		if _, ok := v["__extn"]; ok {
+			return nil, reservedEscapeError(path)
+		}
+
+		keys := make([]string, 0, len(v))
+		for key := range v {
+			keys = append(keys, key)
+		}
+		sort.Strings(keys)
+
+		record := make(cedar.RecordMap, len(v))
+		for _, key := range keys {
+			nested := v[key]
+			converted, err := convertValue(nested, pointer(path, key), depth+1, state, false)
+			if err != nil {
+				return nil, err
+			}
+			if converted != nil {
+				record[cedar.String(key)] = converted
+			}
+		}
+		return cedar.NewRecord(record), nil
+	default:
+		return nil, newConversionError(
+			"invalid_json_value",
+			path,
+			fmt.Sprintf("unsupported json value %T", value),
+		)
+	}
+}
+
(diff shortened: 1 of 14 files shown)


tumberger:
Fixed the evaluator-parity gaps: missing tool input is rejected instead of treated as empty, empty policy sets explicitly default-deny, fixture generation is deterministic, and fixture decoding rejects unknown or trailing data. This keeps enforcement behavior stable and prevents contract changes from passing silently. Unit, race, and vet checks pass.

internal/cedareval/context.go
```diff
@@ -0,0 +1,251 @@
...
+			if err != nil {
+				return nil, err
+			}
+			if converted == nil {
+				return nil, newConversionError(
+					"null_in_set",
+					pointer(path, strconv.Itoa(i)),
+					"null cannot appear in a cedar set",
+				)
+			}
```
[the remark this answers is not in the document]
tumberger: Fixed in 5dbc7b1: removed the unreachable guard. Null array elements already return null_in_set from convertValue before this point.

internal/hookruntime/claude.go
```diff
@@ -46,7 +46,7 @@ type claudeHookSpecificOutput struct {
 
 func DecodeClaudeEvent(input []byte, agentName string) (hook.Event, error) {
 	var h claudeHookInput
-	if err := json.Unmarshal(input, &h); err != nil {
+	if err := decodeUseNumber(input, &h); err != nil {
 		return hook.Event{}, fmt.Errorf("claude: decode hook input: %w", err)
```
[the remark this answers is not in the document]
tumberger: Confirmed: repository-wide audit found no production float64 assertions on ToolInput. UseNumber stays because exact numeric values feed recording, hashing, and Cedar conversion. I also added an EOF check and regression test so the decoder retains json.Unmarshal strictness and rejects trailing JSON.

internal/cedareval/context.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+func reservedEscapeError(path string) error {
+	return newConversionError(
+		"reserved_cedar_escape",
+		path,
+		"raw tool input cannot inject cedar __entity or __extn escape objects",
+	)
+}
+
```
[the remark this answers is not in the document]
tumberger: Fixed in 5dbc7b1: removed the redundant zero branch. Converting positive or negative zero to int64 already produces the same Cedar long.

tumberger:
@greptile review again from first principles

tumberger:
* **#382** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/382?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/382?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

internal/cedareval/evaluator.go
```diff
@@ -0,0 +1,215 @@
...
+func New(policyText string) (*Evaluator, error) {
+	if len([]byte(policyText)) > PolicyMaxBytes {
+		return nil, fmt.Errorf("cedareval: policy text exceeds %d bytes", PolicyMaxBytes)
+	}
+
+	policies, err := cedar.NewPolicySetFromBytes("policy.cedar", []byte(policyText))
+	if err != nil {
+		return nil, fmt.Errorf("cedareval: parse policy: %w", err)
+	}
+	return &Evaluator{policies: policies}, nil
```
tumberger, line 83: Please validate policy metadata here, once at load time: every policy needs a non-empty unique @id, and @ask must be absent or exactly prompt. Today malformed metadata fails only when that policy matches, so one cached evaluator can work for some requests and fail for others.

internal/cedareval/context.go
```diff
@@ -0,0 +1,241 @@
...
+		return nil, newConversionError(
+			"invalid_json_value",
+			path,
+			fmt.Sprintf("unsupported json value %T", value),
+		)
+	}
+}
+
+func convertNumber(number string, path string) (cedar.Value, error) {
+	value, err := strconv.ParseFloat(number, 64)
```
tumberger, line 175: This rounds the exact JSON token preserved by the decoder. For example, 0.00010000000000000001 becomes 0.0001 and Cedar evaluates a different value. Please define the portable numeric semantics across Go and WASM, then add the same long-tail and exponent fixtures to both runtimes. This should not be changed in Go alone.

tumberger: Fixed across both runtimes. Request-contract v1 now explicitly defines JSON numbers as binary64 values, matching JSON.parse; exact lexical or higher-precision inputs must be strings. The shared portable corpus includes the exact long-tail token and exponent spelling, and both TypeScript and Go produce the same Cedar decimal result byte-for-byte.

internal/cedareval/evaluator.go
tumberger: Fixed at evaluator construction. New now rejects missing or empty @id, duplicate stable IDs, and any @ask value other than prompt before the evaluator can be cached. The table-driven tests cover all four invalid metadata cases.

michiosw (commented):
**Verdict: request changes before enforce mode ships** (posted as comments; consolidated multi-reviewer pass — Claude / Codex / Kimi K3, findings hand-verified).

The harness itself is strong — digest-pinned fixtures, strict decoding, `DeepEqual` on full structs, `-race` clean, and the fail-open audit of the decision surface came back negative. One real correctness bug in the decimal converter (inline, empirically reproduced), plus two minor hardening notes.

internal/cedareval/context.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	// Accept exactly the doubles that round-trip through their own four-digit
+	// decimal rendering. Scaling by 10^4 in binary64 injects rounding noise and
+	// makes acceptance depend on luck (0.07 * 10000 != 700 in binary64), and it
+	// collapses distinct doubles into one decimal. Only non-integers reach
+	// here, so the 'f' rendering is exact and bounded. TypeScript mirror uses
+	// Number(value.toFixed(4)) === value; exact-halfway doubles fail the
+	// round-trip in both runtimes, so tie-breaking rules can never disagree.
+	fixed := strconv.FormatFloat(value, 'f', 4, 64)
```
michiosw: **The tie-breaking invariant this comment relies on is false for |v| ≥ ~2^39, producing a cross-runtime allow/deny divergence.**

Once double spacing reaches 1e-4 (non-integer |v| in [2^39, 2^52)), *both* adjacent 4-digit decimals parse back to the same double, so an exact-halfway value passes the round-trip check in both runtimes while each renders a different decimal. Empirically (Go 1.25 / node):

- input `549755813888.03125` (exact double, 2^39 + 2^-5)
- Go `strconv.FormatFloat(v, 'f', 4, 64)` → `"549755813888.0312"` (round-half-even), round-trips → **accepted**
- JS `v.toFixed(4)` → `"549755813888.0313"` (half-up on ties), `Number(s) === v` → **also accepted**

A policy comparing `context.x == decimal("549755813888.0312")` then decides differently in the Go evaluator vs the TS mirror for identical input — under enforce, a live allow/deny disagreement. The portable fixtures (`binary64-json-number-semantics`, `decimal-two-fractional-digits`, `too-many-decimal-places`) all use |v| < 1, so the conformance harness passes while behavior diverges.

Suggested fix: after the round-trip check, reject values whose halfway rendering is ambiguous (e.g. reject when the 5-digit rendering ends in a nonzero 5th digit of exactly 5), or canonically derive digits the same way in both runtimes — and add portable fixtures in the [2^39, 2^52) regime either way.

internal/cedareval/fixtures_test.go
```diff
@@ -0,0 +1,404 @@
...
+	"github.com/kontext-security/kontext-cli/internal/agent"
+	"github.com/kontext-security/kontext-cli/internal/agent/claude"
+	"github.com/kontext-security/kontext-cli/internal/agent/codex"
+	"github.com/kontext-security/kontext-cli/internal/agent/cowork"
+	"github.com/kontext-security/kontext-cli/internal/cedareval"
+)
+
+const portableFixtureContractVersion = 1
+
+var fixtureDigests = map[string]string{
```
michiosw, line 26: Minor (harness soundness): provenance pins only the files listed here — a new fixture file dropped into `testdata/portable/v1/` is neither digest-pinned nor exercised (no directory-listing completeness check), and the per-fixture `version` fields are decoded but never asserted (only the decision-mapping test checks its contract version). A `ReadDir` assertion that every file in the directory appears in `fixtureDigests` would close it cheaply.

internal/cedareval/evaluator.go
```diff
@@ -0,0 +1,253 @@
...
+func annotationValue(annotations cedar.Annotations, name string) (string, bool) {
+	for key, value := range annotations {
+		if string(key) == name {
+			return string(value), true
+		}
+	}
+	return "", false
+}
+
+func validateInput(input ToolUseInput) error {
```
michiosw, line 211: Minor: `EntityID`/`ToolName` are not checked with `utf8.ValidString`, unlike policy IDs (`decision.go`) and the hashing preimages. `stringLength`'s `utf16.Encode([]rune(v))` converts invalid bytes to U+FFFD, so `"\xff"` passes as length 1 and flows into `cedar.String` — an input the UTF-16-native TS runtime can never see. Only reachable by direct Go API callers (JSON decoding sanitizes), but inconsistent with the rest of the contract surface.

tumberger: Fixed in 4fd935f. EntityID and ToolName now require valid UTF-8 before UTF-16 length validation, with direct Go caller regressions.

internal/cedareval/context.go
tumberger: Fixed in 4fd935f. Go now derives the four-digit decimal with the exact ECMAScript toFixed tie rule before the existing round-trip check. Positive and negative 2^39-regime regressions cover the divergence.

internal/cedareval/fixtures_test.go
tumberger: Fixed in 4fd935f. Provenance now scans every JSON fixture in v1 for a pinned digest, and the shared loader asserts the contract version for both array- and object-shaped fixture files.

michiosw (approved):

michiosw:
### Merge activity

* **Jul 22, 7:57 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/382).
* **Jul 22, 7:58 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/382) rebased this pull request as part of a merge.

PR #383, opened 2026-07-14
feat(policy): add Go Cedar decision conformance

tumberger:
## Why

Observe and enforce modes need the same deterministic decision state machine and evidence vocabulary on the endpoint and server.

## What

- mirrors the versioned rollout mapper and reason-code contract in Go
- preserves existing authority in observe mode and fails closed in enforce mode
- rejects contradictory enforce/disabled states
- deduplicates and orders policy evidence by UTF-8 bytes
- pins BMP/non-BMP ordering and all rollout outcomes in portable fixtures

## Verification

- full `go test -race ./...`
- `go vet ./...`
- byte-identical cross-runtime fixture comparison

## Change size

7 files, +1,223/-3 — decision mapper, validation, portable fixtures, and tests.

Tracks ENG-536.

PR #384, opened 2026-07-15
feat(policy): cache native Cedar deployments

tumberger:
## Why

Local Cedar evaluation needs a validated policy mirror that refreshes outside the hook path and remains usable through bounded network failure.

## What

- fetches GET /api/v1/installations/:id/policy with version negotiation and conditional ETags
- accepts a policy-only ready response and strict reduced-v1 states, including principal_unavailable
- validates policy hash, principal, rollout mode, and the seven-term semantic deployment identity
- copies and digest-pins the canonical GetPolicy state fixture
- atomically persists a private last-known-good cache
- restores every retained non-ready state after a matching 304 and expires policy by age
- starts immediate and periodic refresh under daemon lifecycle cancellation

## Scope

Policy retrieval and caching only. No endpoint configuration fields and no hook-decision changes.

Depends on #383. Tracks ENG-533.

## Verification

- go test ./...
- focused go test -race across Cedar policy and daemon packages
- go vet ./...

Change size: 12 files, +1,451/-11 — GetPolicy contract/client/cache, daemon lifecycle, portable fixture, and tests.

### DIFF
--- a/internal/cedarpolicy/cache.go
+++ b/internal/cedarpolicy/cache.go
@@ -353,7 +353,11 @@ func (r *Refresher) Refresh(ctx context.Context) error {
 		r.recordFailure(err)
 		return err
 	}
-	return r.Cache.Apply(result, r.now())
+	if err := r.Cache.Apply(result, r.now()); err != nil {
+		r.recordFailure(err)
+		return err
+	}
+	return nil
 }
 
 func (r *Refresher) Run(ctx context.Context) {
--- a/internal/cedarpolicy/cache.go
+++ b/internal/cedarpolicy/cache.go
@@ -1,0 +1,402 @@
+package cedarpolicy
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+)
+
+const (
+	DefaultRefreshInterval = time.Minute
+	DefaultMaxAge          = 15 * time.Minute
+)
+
+type CacheStatus struct {
+	State         State
+	FetchedAt     time.Time
+	LastAttemptAt time.Time
+	Stale         bool
+	Expired       bool
+	LastError     string
+	Invalid       bool
+}
+
+type Snapshot struct {
+	Deployment    *Deployment
+	LastKnownGood *Deployment
+	State         State
+	Status        CacheStatus
+}
+
+// SnapshotProvider exposes validated in-memory policy state to the hook path.
+// Current never performs filesystem or network I/O.
+type SnapshotProvider interface {
+	Current() Snapshot
+}
+
+type cacheFile struct {
+	Version    int         `json:"version"`
+	State      State       `json:"state"`
+	FetchedAt  string      `json:"fetchedAt"`
+	Deployment *Deployment `json:"deployment,omitempty"`
+	LastGood   *Deployment `json:"lastGood,omitempty"`
+}
+
+type Cache struct {
+	path   string
+	maxAge time.Duration
+	now    func() time.Time
+
+	mu       sync.RWMutex
+	state    State
+	fetched  time.Time
+	active   *Deployment
+	lastGood *Deployment
+	status   CacheStatus
+}
+
+func NewCache(path string, maxAge time.Duration) *Cache {
+	if maxAge <= 0 {
+		maxAge = DefaultMaxAge
+	}
+	return &Cache{path: path, maxAge: maxAge, now: time.Now}
+}
+
+func DefaultCachePathForDB(dbPath string) string {
+	return filepath.Join(filepath.Dir(dbPath), "cedar-policy.json")
+}
+
+func (c *Cache) Load() error {
+	if c.path == "" {
+		return nil
+	}
+	data, err := os.ReadFile(c.path)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		return fmt.Errorf("cedar policy cache: read: %w", err)
+	}
+	var file cacheFile
+	if err := decodeStrict(strings.NewReader(string(data)), &file); err != nil {
+		return fmt.Errorf("cedar policy cache: %w", err)
+	}
+	if file.Version != 1 {
+		return fmt.Errorf("cedar policy cache: unsupported version %d", file.Version)
+	}
+	if file.Deployment != nil {
+		if err := file.Deployment.Validate(); err != nil {
+			return fmt.Errorf("cedar policy cache: invalid deployment: %w", err)
+		}
+	}
+	if file.LastGood != nil {
+		if err := file.LastGood.Validate(); err != nil {
+			return fmt.Errorf("cedar policy cache: invalid last-known-good deployment: %w", err)
+		}
+	}
+	if file.State == StateSuccess && file.Deployment == nil && file.LastGood == nil {
+		return errors.New("cedar policy cache: success state has no deployment")
+	}
+	fetchedAt, err := time.Parse(time.RFC3339Nano, file.FetchedAt)
+	if err != nil {
+		return fmt.Errorf("cedar policy cache: invalid fetchedAt: %w", err)
+	}
+
+	c.mu.Lock()
+	c.state = file.State
+	c.fetched = fetchedAt
+	c.active = cloneDeployment(file.Deployment)
+	c.lastGood = cloneDeployment(file.LastGood)
+	if c.lastGood == nil && file.Deployment != nil {
+		c.lastGood = cloneDeployment(file.Deployment)
+	}
+	c.status = CacheStatus{
+		State:     file.State,
+		FetchedAt: fetchedAt,
+		Stale:     true,
+		LastError: "persisted deployment not yet confirmed",
+	}
+	c.mu.Unlock()
+	return nil
+}
+
+func (c *Cache) Apply(result FetchResult, fetchedAt time.Time) error {
+	if fetchedAt.IsZero() {
+		fetchedAt = c.now().UTC()
+	}
+	c.mu.RLock()
+	file := cacheFile{
+		Version:    1,
+		State:      c.state,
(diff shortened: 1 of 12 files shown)


tumberger:
* **#384** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/384?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/384?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

internal/cedarpolicy/cache.go
```diff
@@ -0,0 +1,402 @@
...
+	}
+	data, err := os.ReadFile(c.path)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		return fmt.Errorf("cedar policy cache: read: %w", err)
+	}
+	var file cacheFile
+	if err := decodeStrict(strings.NewReader(string(data)), &file); err != nil {
```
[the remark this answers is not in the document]
tumberger: One-time load of a small cache file; the extra copy is negligible. Resolving as noise.

internal/cedarpolicy/cache.go
```diff
@@ -0,0 +1,402 @@
...
+		LastAttemptAt: fetchedAt,
+	}
+	c.mu.Unlock()
+	return nil
+}
+
+func (c *Cache) MarkFailed(err error, attemptedAt time.Time) {
+	if attemptedAt.IsZero() {
+		attemptedAt = c.now().UTC()
+	}
```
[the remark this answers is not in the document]
tumberger: Acknowledged, low-priority: the read-lock/write-lock gap is real, but Apply runs only from the serial Refresher loop, so there is no concurrent caller today. Noting it here and resolving.

internal/cedarpolicy/client.go
```diff
@@ -0,0 +1,154 @@
...
+			}
+			if etag == "" || etag != deployment.DeploymentIdentity {
+				return FetchResult{}, errors.New("cedar policy: ETag does not match deployment identity")
+			}
+			return FetchResult{State: StateSuccess, Deployment: &deployment, ETag: etag}, nil
+		}
+		return decodeStateResult(body, etag)
+	}
+
+	if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusNotAcceptable || resp.StatusCode == http.StatusServiceUnavailable {
```
michiosw, line 121: **Important: no status↔state pairing — a 503 carrying a valid `disabled` body is applied as an explicit rollback.**

All three statuses (401/406/503) funnel into `decodeStateResult` without checking that the decoded state is one this status may legitimately carry. A backend or proxy contract bug that returns 503 with a valid `{"state":"disabled"}` (or `no_active_policy`) body gets applied as an explicit rollback — Cedar authority relinquished, legacy evaluator restored — instead of being treated as a transient failure. Suggest enforcing the pairing (e.g. 503 may only carry retryable/unavailable states; explicit rollback states only on their designated status).

internal/cedarpolicy/contract.go
```diff
@@ -0,0 +1,209 @@
...
+	if d.RolloutMode != cedareval.RolloutModeObserve && d.RolloutMode != cedareval.RolloutModeEnforce {
+		return fmt.Errorf("cedar policy: unsupported rollout mode %q", d.RolloutMode)
+	}
+	principalLength := utf16Length(d.EvaluationPrincipal.EntityID)
+	if d.EvaluationPrincipal.EntityType != cedareval.PrincipalEntityType || principalLength == 0 || principalLength > 1024 {
+		return errors.New("cedar policy: invalid evaluation principal")
+	}
+	if d.Signature != "" {
+		signatureLength := utf16Length(d.Signature)
+		if signatureLength == 0 || signatureLength > 8192 {
```
michiosw, line 56: **Important: `Signature` is length-checked here but never verified anywhere in the repo.**

`Deployment.Validate()` is otherwise self-referential (recomputes `PolicyHash` from the file's own `PolicyText`, `DeploymentIdentity` from its own metadata), so cache integrity is entirely self-certifying. `managed.json` is MDM-controlled, but `cedar-policy.json` is same-user-writable (0600 under the user's own dir): a same-UID process can forge a fully "valid" file with `state:"disabled"` or `rolloutMode:"observe"` and downgrade enforcement until the next refresh (up to maxAge offline). Either wire up actual signature verification or explicitly document the same-user trust boundary — shipping the dead field suggests a guarantee that doesn't exist.

internal/cedarpolicy/cache.go
```diff
@@ -0,0 +1,402 @@
...
+	if err != nil {
+		r.recordFailure(err)
+		return err
+	}
+	if result.State == StateUnavailable {
+		err = errors.New("cedar policy is temporarily unavailable")
+		r.recordFailure(err)
+		return err
+	}
+	return r.Cache.Apply(result, r.now())
```
michiosw: Minor (observability with real teeth): every other failure path in `Refresh` records the failure, but an `Apply` error (persist failure on disk-full/unwritable dir, rejected result) is returned without `recordFailure` — backoff triggers, yet `Status.Stale` stays false and `LastError` empty. Hook evidence then reports `cacheStale: false` while the old deployment keeps serving, and a fetched observe→enforce cutover can sit unapplied until maxAge expiry. The `endpointconfig` sibling records these (`r.fail(err)`); suggest aligning.

internal/cedarpolicy/cache.go
```diff
@@ -0,0 +1,402 @@
...
+			return err
+		}
+		file.State = StateSuccess
+		file.Deployment = cloneDeployment(result.Deployment)
+		file.LastGood = nil
+	case StateNotModified:
+		if file.LastGood == nil || result.ETag == "" || result.ETag != file.LastGood.DeploymentIdentity {
+			return errors.New("cedar policy cache: not-modified result does not match last-known-good deployment")
+		}
+		file.State = StateSuccess
```
michiosw, line 158: Minor (confirm intent): after an explicit `no_active_policy`/`unauthorized`/`disabled` state, `lastGood` is retained and `ConditionalIdentity()` keeps advertising the retired identity — so a later 304 resurrects the withdrawn policy with a fresh `fetchedAt`, without the server ever re-sending it (`TestCacheExplicitStateThenMatchingNotModifiedRestoresPolicy` asserts this deliberately). A buggy server ETag comparison or an intermediary answering the conditional GET becomes a policy-resurrection channel. If intended, worth a comment saying why; if not, drop the ETag after explicit rollback states.

internal/cedarpolicy/cache.go
```diff
@@ -0,0 +1,402 @@
...
+	fetched := c.fetched
+	active := cloneDeployment(c.active)
+	lastGood := cloneDeployment(c.lastGood)
+	status := c.status
+	c.mu.RUnlock()
+
+	if active == nil && state == StateSuccess {
+		active = lastGood
+	}
+	if !fetched.IsZero() && now.Sub(fetched) > c.maxAge {
```
michiosw, line 236: Minor: expiry runs on wall-clock time — `fetched` is parsed from disk (RFC3339) or `.UTC()`-stripped, so there's no monotonic reading. A backwards clock step (NTP correction, manual change) extends a stale/revoked policy's serving window arbitrarily, weakening the `!Expired` gate enforcement relies on. Low exploitability; noting since this cache feeds an enforcement decision.

michiosw (commented):
**Verdict: approve with comments** (posted as comments; consolidated multi-reviewer pass — Claude / Codex / Kimi K3, findings hand-verified).

Carefully built cache: validate-before-install, properly atomic persisted writes (temp + fsync + rename + dir-fsync, 0600), fail-closed corruption handling, tight wire-contract validation. Nothing merge-blocking, but two of the inline items (status↔state pairing, unverified `Signature`) deserve a deliberate decision before the enforce cutover relies on this cache.

internal/cedarpolicy/cache.go
tumberger: Valid low-exploitability hardening point. A persisted RFC3339 timestamp cannot retain a monotonic component, so handling backward wall-clock movement needs an explicit skew policy. Deferring that cache-hardening choice; it is not a release blocker for this stack.

internal/cedarpolicy/cache.go
tumberger: Fixed in 58fc9ab. Refresher now records Cache.Apply failures through the same failure path as token and fetch errors, including stale status, LastError, and LastAttemptAt. A persistence-failure regression covers it.

internal/cedarpolicy/client.go
tumberger: Fixed in 58fc9ab. The client now enforces the exact HTTP status to state contract: 200 only explicit non-deployment states, 401 unauthorized, 406 unsupported_version, and 503 unavailable. Mismatch regressions cover each status family.

internal/cedarpolicy/cache.go
tumberger: Intentional for v1: lastGood and its conditional identity are retained so the same deployment can recover through a matching 304. The client now strictly validates status and state pairing; dropping conditional recovery after every explicit state would require a separate server contract change.

internal/cedarpolicy/contract.go
tumberger: Agreed that this is a real trust-boundary limitation. This release does not claim resistance to same-UID cache tampering, and verifying only the optional Signature field would still leave explicit cached states unsigned. Deferring signing to a complete cache-envelope design instead of adding partial verification here.

michiosw (approved):

michiosw:
### Merge activity

* **Jul 22, 7:57 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/384).
* **Jul 22, 8:07 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/384) rebased this pull request as part of a merge.

PR #386, opened 2026-07-15
feat(policy): gate Cedar enforcement cutover

tumberger:
## Why

Cedar enforcement needs a singular authority boundary with deterministic behavior during startup, refresh failure, expiration, identity failure, incompatibility, corruption, and explicit rollback.

## What

- grants Cedar authority only when the local cutover gate is enabled and the semantic deployment requests enforce
- stops calling the previous evaluator after cutover
- continues an age-valid last-known-good deployment through transient refresh failure
- denies when enforcement has no trusted ready deployment, including no-LKG, principal_unavailable, unauthorized, unsupported, expired, and invalid states
- returns authority only for configured observe, disabled, or no_active_policy
- maps ASK to deny until an approval channel exists
- records distribution state and cache readiness in decision evidence

## Scope

Authorization cutover only. Endpoint configuration remains independent and cannot grant or revoke policy authority.

Depends on #385. Tracks ENG-535.

## Verification

- go test ./...
- focused go test -race across policy, server, ledger, runtime, and daemon packages
- go vet ./...

Change size: 8 files, +296/-21 — authority/readiness gate, fail-closed evidence, and transition tests.

### DIFF
--- a/internal/guard/app/server/cedar.go
+++ b/internal/guard/app/server/cedar.go
@@ -129,14 +129,21 @@ func (p *cedarPolicyProvider) evaluate(snapshot cedarpolicy.Snapshot, event risk
 				}
 				evidence.EngineErrorCount = 1
 			} else {
-				outcome = cedareval.EvaluationOutcome{
-					State:                cedareval.EvaluationStateEvaluated,
-					Decision:             result.Decision,
-					Ask:                  result.Ask,
-					DeterminingPolicyIDs: result.DeterminingPolicyIDs,
-				}
 				evidence.ContextDiagnostics = result.ContextDiagnostics
 				evidence.EngineErrorCount = len(result.EngineDiagnostics.Errors)
+				if evidence.EngineErrorCount > 0 {
+					outcome = cedareval.EvaluationOutcome{
+						State:  cedareval.EvaluationStateFailed,
+						Reason: cedareval.ReasonEngineError,
+					}
+				} else {
+					outcome = cedareval.EvaluationOutcome{
+						State:                cedareval.EvaluationStateEvaluated,
+						Decision:             result.Decision,
+						Ask:                  result.Ask,
+						DeterminingPolicyIDs: result.DeterminingPolicyIDs,
+					}
+				}
 			}
 		}
 	} else if snapshot.Status.Invalid {
--- a/internal/guard/app/server/cedar.go
+++ b/internal/guard/app/server/cedar.go
@@ -14,8 +14,9 @@ import (
 const cedarEvaluatorVersion = "cedar-go/v1.8.0"
 
 type cedarPolicyProvider struct {
-	current   PolicyProvider
-	snapshots cedarpolicy.SnapshotProvider
+	current            PolicyProvider
+	snapshots          cedarpolicy.SnapshotProvider
+	enforcementEnabled bool
 
 	mu        sync.Mutex
 	identity  string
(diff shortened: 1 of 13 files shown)


tumberger:
* **#386** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/386?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/386?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tumberger:
## Gap: enforce decisions can't reach the agent through the managed-observe hook path

This PR grants the daemon enforce authority (`CedarEnforcement`, `claimsAuthority`, and `clientResultTransform` returning nil for enforce so a deny passes through the socket unmodified). But the hook binary on the other side of that socket always overrides the result:

- `managedobserve.Lifecycle` routes every result through `observeResult` (`internal/managedobserve/lifecycle.go:137-153`), which unconditionally sets `result.Decision = hook.DecisionAllow` and `result.Mode = observe` — a daemon `deny` is rewritten to `allow` before it returns to the agent.
- `shouldUseManagedObserve` (`cmd/kontext/main.go:331`) selects that path whenever managed config exists and no explicit `--mode` flag is passed, without consulting the configured mode, so an org set to enforce still takes the observe wrapper.

Net effect: with this PR, the daemon evaluates and records enforce correctly, but no deny is delivered to the agent through the production managed path. If that's intentional staging (enforce delivery lands in a follow-up), a one-line note here would make it explicit; otherwise the lifecycle wrapper needs to honor the configured mode and let a daemon deny pass through. Either way, an end-to-end test that asserts "daemon deny → agent sees deny under enforce" would lock the contract — the current gate exercises the daemon side only.

internal/guard/app/server/cedar.go
```diff
@@ -89,20 +139,47 @@ func (p *cedarPolicyProvider) evaluate(snapshot cedarpolicy.Snapshot, event risk
...
+			fallback, _ := cedareval.MapDecision(cedareval.DecisionMappingInput{
+				RolloutMode:      cedareval.RolloutModeEnforce,
+				EnforcementReady: false,
+				Evaluation:       cedareval.EvaluationOutcome{State: cedareval.EvaluationStateFailed, Reason: cedareval.ReasonEngineError},
+			})
+			evidence.Mapping = fallback
+			evidence.AppliedRolloutMode = cedareval.RolloutModeEnforce
+			evidence.EngineErrorCount++
+			return evidence
+		}
```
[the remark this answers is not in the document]
tumberger: Fixed — the enforce fallback now uses the valid ready-but-failed input that maps to a deny, and no longer discards the mapping error (explicit deny if it ever fails to map). Fails closed.

tumberger:
## Fixed: enforce decisions now reach the agent (managed hook wrapper no longer forces allow)

Local end-to-end testing reproduced the enforce-delivery gap: the daemon correctly computes an authoritative `deny` in enforce mode, but the managed hook wrapper (`managedobserve/lifecycle.go:observeResult`) unconditionally rewrote every result to `allow` — so no Cedar policy could ever block a tool call through the managed path.

### Root cause
The guard server already applies the correct per-mode transform (`clientResultTransform`): observe turns a deny into `allow` + "would deny", enforce passes the real decision through. The hook-binary wrapper then re-applied the observe transform a second time, discarding the enforce deny.

### Fix
`Lifecycle` is now mode-aware (reads the managed rollout posture; empty = observe):

- **observe** — unchanged: the wrapper still forces `allow` on any decision (defense-in-depth: observe can never block, even if the daemon didn't transform).
- **enforce** — the daemon's decision passes through unchanged (only normalized), so an authoritative `deny` reaches the agent.
- **daemon unavailable** — observe fails open as before; **enforce fails closed** (deny with `Kontext enforce: managed policy daemon unavailable`) rather than silently allowing. This is the deliberate enforcement posture — an enforce endpoint with no reachable daemon blocks blocking hooks until the daemon is back. Flagging explicitly for review since it has an operational implication (a missing/crashed daemon blocks tool calls under enforce).

### Tests
Existing observe tests unchanged and green (Mode defaults to observe). Added: `TestLifecycleEnforcePassesDaemonDenyThroughToAgent` (daemon deny reaches the agent) and `TestLifecycleEnforceFailsClosedWhenDaemonUnavailable`. Full `managedobserve`, `runtimehost`, `guard/...`, and `cmd/...` suites pass.

Note: this only fixes delivery through the managed-observe hook path. The daemon-side decision was already correct.

tumberger:
Stack housekeeping completed: the two diagnostic-only Cedar benchmark files were removed completely and were not submitted in the restacked CLI branches. No Go Benchmark functions remain in this stack; acceptance evidence is the portable fixture corpus, full go test -race ./..., go vet ./..., Guard smoke/local E2E, and green CI.

michiosw (commented):
**Verdict: request changes** (posted as comments; consolidated multi-reviewer pass — Claude / Codex / Kimi K3, findings hand-verified against the stack tip).

The in-process enforcement core is solid: every traced return path in `claimsAuthority`/`evaluate`/`MapDecision` fails closed, and the erroring-forbid trap is explicitly neutralized (`erroring-forbid-cannot-authorize` fixture + `TestCedarEnforceFailsClosedOnEngineDiagnostics`). The two blockers are at the transport seam and the ledger:

1. Enforce fails open when the daemon RPC errors after a successful probe (inline at `lifecycle.go`).
2. Enforced decisions produce two authoritative `request.decided` rows (inline at `store/sqlite/cedar.go`).

Also inline: the empty-decision default direction, and an operational note on enforce-mode behavior for non-blocking hooks / fresh installs.

internal/managedobserve/lifecycle.go
```diff
@@ -81,18 +105,54 @@ func (l Lifecycle) processWithKickstart(ctx context.Context, event hook.Event, b
...
+	return l.finalize(event, l.call(ctx, event))
+}
+
+// finalize applies the client-side posture to a daemon result. In enforce the
+// daemon already produced the authoritative decision (the guard server does not
+// observe-transform in enforce), so it passes through unchanged — only
+// normalized. In observe the hook must never block, so the decision is forced
+// to allow with a "would" note. This is the boundary that must not silently
+// downgrade an enforce deny to allow.
+func (l Lifecycle) finalize(event hook.Event, result hook.Result) hook.Result {
```
michiosw, line 135: **Blocker — enforce fails open on daemon RPC error after a successful probe.**

`call()` (line 183, unchanged by this PR) returns `hook.Result{Decision: DecisionAllow, Reason: "managed observe daemon unavailable"}` on *any* `client.Process` error — timeout against the PreToolUse budget, connection reset, daemon crash mid-request, truncated/malformed response. Since the socket probe already succeeded, `daemonUnavailable()` is bypassed, and `finalize` here passes the non-empty allow through unchanged in enforce mode. A daemon that accepts connections but is wedged (SQLite stall, slow disk) therefore silently allows tool calls Cedar would deny — exactly what this function's own comment says must not happen. `TestLifecycleEnforceFailsClosedWhenDaemonUnavailable` only exercises probe failure, not call failure.

Related gap in the same boundary: the daemon parses its mode **once at startup** (`daemon.go:130`; its later `managedconfig.Load()` calls are token-only) while this wrapper re-reads `managed.json` per hook, and nothing restarts the daemon on a mode change. After an observe→enforce flip, the stale observe daemon's result-transform rewrites every deny to allow and `finalize` passes it through (`result.Mode == "observe"` is accepted; only empty gets defaulted). Every Cedar deny becomes an allow until the daemon happens to restart.

Suggested fix: in enforce, treat RPC errors, empty/invalid decisions, **and `result.Mode != enforce`** as unavailable → deny (route through `daemonUnavailable`, and consider kickstarting the daemon on mode mismatch). Regression tests: (a) socket accepts the probe but drops/truncates the request; (b) daemon replies `Mode: "observe"` with an allow.

internal/managedobserve/lifecycle.go
```diff
@@ -81,18 +105,54 @@ func (l Lifecycle) processWithKickstart(ctx context.Context, event hook.Event, b
...
+	if l.enforcing() {
+		if result.Decision == "" {
```
michiosw: The empty-decision default in the **enforce** branch is allow — the wrong direction for this posture. It's not clearly reachable today (the daemon stamps allow/deny for PreToolUse and decode failures map to deny for blocking hooks), but if an unexpected empty decision ever arrives here, enforce should deny, consistent with `daemonUnavailable` below.

internal/managedobserve/lifecycle.go
```diff
@@ -81,18 +105,54 @@ func (l Lifecycle) processWithKickstart(ctx context.Context, event hook.Event, b
...
+	return observeResult(event, result)
+}
+
+// daemonUnavailable is the fail path when the managed daemon cannot be reached.
+// Observe fails open (it never blocks). Enforce fails closed: enforcement
+// requires an authoritative decision and an unreachable daemon cannot provide
+// one, so the tool call is denied rather than silently allowed. This means an
+// enforce endpoint with no reachable daemon blocks blocking hooks until the
+// daemon is back — the deliberate fail-closed posture for enforcement.
+func (l Lifecycle) daemonUnavailable(event hook.Event) hook.Result {
```
michiosw, line 153: Non-blocking operational note: this denies *all* hook events when the daemon is down in enforce — including non-blocking ones (PostToolUse, SessionEnd) where a deny is at best noise — and a fresh enforce install denies everything until the first successful policy fetch. If that's the intended cutover posture (it reads deliberate), worth documenting in the rollout notes so support isn't surprised.

internal/guard/store/sqlite/cedar.go
```diff
@@ -74,7 +77,15 @@ func cedarDecisionActionValues(actionID, sessionID string, event risk.HookEvent,
 	}
 }
 
+// cedarDecisionCategory marks whether a Cedar row is an authoritative decision
+// or an observe-mode shadow. In enforce mode Cedar is the authority, so the row
+// is a real policy decision ("cedar_policy", matching the DecisionStage stamped
+// by applyCedarDecision); otherwise it is a dry-run shadow the dashboard filters
+// out of authoritative views.
 func cedarDecisionCategory(evidence risk.CedarEvidence) string {
+	if evidence.AppliedRolloutMode == cedareval.RolloutModeEnforce {
```
michiosw: **Blocker — enforced decisions are recorded twice as authoritative.**

In enforce, `SaveDecision` writes the primary `request.decided` row whose `decision_category` resolves to `cedar_policy` via `decisionCategory(riskEvent)` → `DecisionStage` (stamped by `applyCedarDecision`), and then `insertCedarDecisionAction` writes this auxiliary row — also `canonical_event_type = request.decided`, and with this change also `cedar_policy`. One tool call → two authoritative decided rows: action and denial counts double in local summaries and any hosted query that filters out `dry_run`.

Either fold the Cedar evidence into the primary row and skip the auxiliary insert when Cedar is authoritative, or keep the auxiliary row in a category authoritative queries exclude (e.g. `cedar_evidence`).

tumberger: Fixed in 119ba09. The primary request.decided row remains the single authoritative cedar_policy decision. The dedicated Cedar row is now always evidence-only under dry_run, which preserves evaluator evidence while keeping authoritative counts at exactly one. Regression added.

internal/managedobserve/lifecycle.go
tumberger: Fixed in 119ba09. RPC errors now route through daemonUnavailable, enforce accepts only explicit allow or deny responses stamped with Mode enforce, and stale observe-mode replies fail closed. Regressions cover a socket that accepts probes then drops the RPC and a stale observe response.

michiosw (approved):

michiosw:
### Merge activity

* **Jul 22, 7:57 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/386).
* **Jul 22, 8:20 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/386) rebased this pull request as part of a merge.
* **Jul 22, 8:22 PM UTC**: @michiosw merged this pull request with [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/386).

PR #387, opened 2026-07-15
feat(config): sync endpoint configuration independently

tumberger:
## Why

Payload capture is endpoint behavior, not policy deployment metadata. It needs an independent contract and failure posture so configuration availability cannot change policy authority.

## What

- fetches the versioned installation configuration endpoint with conditional ETags
- validates and digest-pins the portable configuration identity contract
- persists an independent private cache while keeping summary mode effective until a fresh 200 or matching 304 confirms it
- falls back to summary on refresh failure with independent bounded backoff and jitter
- removes payload-capture authority from provider-policy refresh
- snapshots configured mode, effective mode, identity, confirmation, and fallback reason atomically in decision evidence

## Scope

This PR changes endpoint payload-capture configuration only. It does not retrieve Cedar policy, evaluate policy, or change authorization authority.

Depends on #384. Tracks ENG-539.

## Verification

- go test ./...
- focused go test -race across endpoint configuration, policy, server, ledger, and daemon packages
- go vet ./...

Change size: 17 files, +1,235/-67 — configuration contract/client/cache, daemon application, atomic evidence, and tests.

### DIFF
--- a/internal/endpointconfig/cache.go
+++ b/internal/endpointconfig/cache.go
@@ -1,0 +1,375 @@
+package endpointconfig
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"math/rand/v2"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/kontext-security/kontext-cli/internal/payloadcapture"
+)
+
+const (
+	DefaultRefreshInterval = time.Minute
+	DefaultMaxBackoff      = 10 * time.Minute
+)
+
+var defaultConfig = Config{PayloadCaptureMode: payloadcapture.ModeSummary}
+
+type Status struct {
+	FetchedAt     time.Time
+	LastAttemptAt time.Time
+	Stale         bool
+	LastError     string
+	Invalid       bool
+}
+
+type Snapshot struct {
+	Config         Config
+	Configured     Config
+	ConfigIdentity string
+	Confirmed      bool
+	FallbackReason string
+	LastKnownGood  *Response
+	Status         Status
+}
+
+type cacheFile struct {
+	Version   int       `json:"version"`
+	FetchedAt string    `json:"fetchedAt"`
+	Response  *Response `json:"response"`
+}
+
+type Cache struct {
+	path string
+	now  func() time.Time
+
+	mu       sync.RWMutex
+	fetched  time.Time
+	active   *Response
+	lastGood *Response
+	status   Status
+}
+
+func NewCache(path string) *Cache {
+	return &Cache{path: path, now: time.Now}
+}
+
+func DefaultCachePathForDB(dbPath string) string {
+	return filepath.Join(filepath.Dir(dbPath), "endpoint-config.json")
+}
+
+// Load restores only the conditional value. The effective configuration stays
+// at the privacy-safe default until the server confirms it with 200 or 304.
+func (c *Cache) Load() error {
+	if c.path == "" {
+		return nil
+	}
+	data, err := os.ReadFile(c.path)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		return fmt.Errorf("endpoint configuration cache: read: %w", err)
+	}
+	var file cacheFile
+	if err := decodeStrict(strings.NewReader(string(data)), &file); err != nil {
+		return fmt.Errorf("endpoint configuration cache: %w", err)
+	}
+	if file.Version != 1 || file.Response == nil {
+		return errors.New("endpoint configuration cache: invalid cache shape")
+	}
+	if err := file.Response.Validate(); err != nil {
+		return fmt.Errorf("endpoint configuration cache: %w", err)
+	}
+	fetchedAt, err := time.Parse(time.RFC3339Nano, file.FetchedAt)
+	if err != nil {
+		return fmt.Errorf("endpoint configuration cache: invalid fetchedAt: %w", err)
+	}
+	c.mu.Lock()
+	c.fetched = fetchedAt
+	c.active = nil
+	c.lastGood = cloneResponse(file.Response)
+	c.status = Status{FetchedAt: fetchedAt, Stale: true, LastError: "persisted configuration not yet confirmed"}
+	c.mu.Unlock()
+	return nil
+}
+
+func (c *Cache) Apply(result FetchResult, fetchedAt time.Time) error {
+	if fetchedAt.IsZero() {
+		fetchedAt = c.now().UTC()
+	}
+	c.mu.RLock()
+	lastGood := cloneResponse(c.lastGood)
+	c.mu.RUnlock()
+	var confirmed *Response
+	switch {
+	case result.NotModified:
+		if lastGood == nil || result.ETag == "" || result.ETag != lastGood.ConfigIdentity {
+			return errors.New("endpoint configuration cache: not-modified result does not match last-known-good config")
+		}
+		confirmed = lastGood
+	case result.Response != nil:
+		if err := result.Response.Validate(); err != nil {
+			return err
+		}
+		confirmed = cloneResponse(result.Response)
+	default:
+		return errors.New("endpoint configuration cache: result has no configuration")
+	}
+	file := cacheFile{
+		Version:   1,
+		FetchedAt: fetchedAt.UTC().Format(time.RFC3339Nano),
+		Response:  cloneResponse(confirmed),
+	}
+	if err := c.persist(file); err != nil {
+		return err
+	}
+	c.mu.Lock()
+	c.fetched = fetchedAt
+	c.active = cloneResponse(confirmed)
+	c.lastGood = cloneResponse(confirmed)
+	c.status = Status{FetchedAt: fetchedAt, LastAttemptAt: fetchedAt}
+	c.mu.Unlock()
+	return nil
+}
+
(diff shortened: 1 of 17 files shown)


tumberger:
* **#387** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/387?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/387?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

internal/endpointconfig/cache.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+	c.mu.Unlock()
+}
+
+func (c *Cache) MarkInvalid(err error) {
+	c.MarkFailed(err, c.now().UTC())
+	c.mu.Lock()
+	c.status.Invalid = true
+	c.mu.Unlock()
+}
```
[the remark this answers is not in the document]
tumberger: Fixed in 286971f0 — MarkInvalid now performs failure-state reset and Invalid stamping in one mutex critical section through markFailedLocked. A concurrent Apply versus MarkInvalid regression test asserts no confirmed snapshot can carry Invalid, and the focused race suite plus full CLI CI pass.

internal/endpointconfig/cache.go
```diff
@@ -0,0 +1,375 @@
...
+	if err := directory.Sync(); err != nil {
+		return fmt.Errorf("endpoint configuration cache: sync directory: %w", err)
+	}
+	return nil
+}
+
+func cloneResponse(input *Response) *Response {
+	if input == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
tumberger: Documented and tested in 286971f0 and 98f19068 — after rename succeeds but parent-directory sync fails, Apply intentionally reports failure and does not activate the new value in memory. A surviving renamed file is unconfirmed after restart until a valid 200 or matching 304; the next refresh heals the extra round trip. Cache race tests and CI pass.

internal/managedobserve/daemon.go
```diff
@@ -172,19 +189,15 @@ func RunDaemon(ctx context.Context, opts DaemonOptions) error {
...
-	// Restore the capture mode from the persisted snapshot before the first
-	// fetch completes, so an offline restart keeps the org's last-known
-	// directive instead of silently reverting to the "summary" default.
-	if snapshot, _, ok := githubCache.CurrentSnapshot(); ok {
-		host.SetPayloadCaptureMode(payloadcapture.NormalizeMode(snapshot.PayloadCaptureMode))
-	}
+	// Persisted endpoint configuration is conditional state only. Capture stays
+	// at the privacy-safe default until this process receives a confirming 200
+	// or matching 304 from the independent configuration endpoint.
+	host.SetPayloadCaptureConfiguration(captureConfiguration(endpointConfigCache.Current()))
```
michiosw, line 195: **Important — pre-merge check: server deploy-order dependency.**

The capture directive's *only* source is now `GET /installations/{id}/configuration` — the github-policy-snapshot path and its offline-restart restore ("so an offline restart keeps the org's last-known directive instead of silently reverting to the 'summary' default") were removed. If a CLI carrying this change reaches endpoints before the server route is live, every refresh fails and existing full-capture orgs are silently pinned to summary capture indefinitely. Privacy-safe direction, but a real behavioral regression — please confirm the server route ships first (or restore the persisted-snapshot fallback until it has). `providerpolicy/snapshot.go`'s `PayloadCaptureMode` is now vestigial either way and can be removed once this settles.

internal/endpointconfig/cache.go
```diff
@@ -0,0 +1,384 @@
...
+		confirmed = cloneResponse(result.Response)
+	default:
+		return errors.New("endpoint configuration cache: result has no configuration")
+	}
+	file := cacheFile{
+		Version:   1,
+		FetchedAt: fetchedAt.UTC().Format(time.RFC3339Nano),
+		Response:  cloneResponse(confirmed),
+	}
+	if err := c.persist(file); err != nil {
```
michiosw, line 131: Minor: `Apply` persists unconditionally — including on 304s where only `FetchedAt` changed — and `persist` does file-fsync + dir-fsync. At the 1-minute default interval that's ~2 fsyncs/min forever on every managed endpoint, buying nothing across restarts (the restored value is unconfirmed regardless of `FetchedAt`). Skipping persist when the confirmed identity is unchanged removes the steady-state churn.

internal/endpointconfig/cache.go
```diff
@@ -0,0 +1,384 @@
...
+func (c *Cache) MarkFailed(err error, attemptedAt time.Time) {
+	if attemptedAt.IsZero() {
+		attemptedAt = c.now().UTC()
+	}
+	c.mu.Lock()
+	c.markFailedLocked(err, attemptedAt)
+	c.mu.Unlock()
+}
+
+func (c *Cache) markFailedLocked(err error, attemptedAt time.Time) {
```
michiosw, line 154: Minor: two behaviors here compound on transient failures. (1) Any single error — token load, network blip, one 503 — nils `active` immediately, dropping effective capture from full to summary at once (documented intent, but it makes capture mode and the per-record `confirmed`/`fallback_reason` evidence flap on every blip). (2) `markFailedLocked` never clears `status.Invalid`, so after one corrupt-cache start, every later *network* failure reports `fallback_reason: "invalid_cache"` until a refresh succeeds — misleading diagnostics.

internal/endpointconfig/cache.go
```diff
@@ -0,0 +1,384 @@
...
+	}
+	delay := interval
+	for {
+		if ctx.Err() != nil {
+			return
+		}
+		if err := r.Refresh(ctx); err == nil {
+			delay = interval
+		} else if delay < maxBackoff {
+			delay *= 2
```
michiosw, line 339: Minor: the very first failure already takes the `delay *= 2` branch, so one 503 costs a jittered ~2× interval (~1.5–2.5 min at defaults) of degraded capture before the first retry. Retrying once at the base interval before doubling would make single-blip recovery cheap. (Also, `maxBackoff < interval` at line 325-329 silently substitutes `10 * interval` — a caller passing `Interval: 1h, MaxBackoff: 30m` gets 10h; test-knob edge, but surprising.)

michiosw (commented):
**Verdict: approve with comments** (posted as comments; consolidated multi-reviewer pass — Claude / Codex / Kimi K3, findings hand-verified).

The sync loop, shutdown handling, and in-memory atomicity are solid and well-tested (`-race` clean), and the privacy fail-safe direction is consistently enforced. One pre-merge check (deploy order, inline at `daemon.go`), plus a dead-code deletion that can't be anchored inline:

**Dead configuration path worth deleting:** the legacy `SetPayloadCaptureMode` forwarding chain — `store/sqlite/store.go:37`, `runtimehost/host.go`, `guard/app/server/server.go` — has no production callers left now that everything goes through atomic `SetPayloadCaptureConfiguration`. It also fabricates an unconfirmed `RuntimeConfiguration{EffectiveMode: full, Confirmed: false}`, which is exactly the state `SafeRuntimeConfiguration` exists to forbid. Removing it leaves one configuration path and avoids future drift.

internal/endpointconfig/cache.go
tumberger: Valid steady-state optimization, but deferred for this release. The current write path prioritizes one durable implementation for all successful refreshes; skipping unchanged-identity persistence can be added independently without changing the configuration contract.

internal/managedobserve/daemon.go
tumberger: Confirmed: the server source dependency in kontext PR #685 is merged, while CLI PR #387 is still unmerged, so source-level release order is satisfied. Production deployment ordering remains a release check rather than a code change in this PR.

internal/endpointconfig/cache.go
tumberger: Valid retry-tuning point, but not release-critical. Production uses the bounded defaults; changing first-failure timing and the MaxBackoff override semantics is deferred to a focused refresher policy change.

internal/endpointconfig/cache.go
tumberger: Both behaviors are intentional privacy-safe defaults. Any unconfirmed refresh immediately falls back to summary, and Invalid remains set until a successful response actually revalidates the cache. Clearing it on another failure would imply validation that did not occur.

tumberger:
Also addressed the unanchored cleanup note in d12c05a: removed the dead SetPayloadCaptureMode forwarding chain from Host to Server to Store. Tests now use the full RuntimeConfiguration path. providerpolicy.PayloadCaptureMode remains only as a wire-compatible decode field for older snapshot responses; endpoint configuration is the sole active source.

michiosw (approved):

michiosw:
### Merge activity

* **Jul 22, 7:57 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/387).
* **Jul 22, 8:12 PM UTC**: [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/387) rebased this pull request as part of a merge.
* **Jul 22, 8:14 PM UTC**: @michiosw merged this pull request with [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/387).

PR #392, opened 2026-07-20
fix(sqlite): chunk ledger IN() queries and guard wide receipt rowid ranges

michiosw:
## Summary

The managed-observe ledger export built single `IN (...)` queries with one SQL placeholder per id. On a live 749 MB guard.db with long interleaved agent sessions, the receipt rowid range for one 100-action export page referenced **34,814 distinct action ids** — past modernc.org/sqlite's 32,766-variable cap — so `LedgerBatch` errored on every flush, *after* reading and JSON-decoding ~50 MB of receipts and *before* any cursor advance. The daemon retried this every 10s: export cursor frozen for 8 days, ~46% steady CPU, and the pre-wedge re-send pattern behind the ENG-546 staging aggregation incident.

- Chunk every unbounded `IN ()` (actions by id, agent sessions, receipt bounds) at 500 params, merge-sorting chunk results in Go to preserve the export ordering.
- Add `ReceiptLimit` to `LedgerExportOptions`: count the receipt rowid range **before** materializing it and return `ErrLedgerReceiptRangeTooLarge` — carrying the actions page and cursor so the caller can advance past a poison batch — instead of decoding payloads that can never ship. Wide ranges happen when action update-order and receipt insert-order diverge (interleaved sessions).

## Verification

- [x] `go test ./...` (two pre-existing timing-flaky llama-server tests pass in isolation)
- [x] `go vet` on affected packages
- [x] Regression test: `authorizationActionsByIDs` with 33,000 ids (fails with "too many SQL variables" before this change)
- [x] Replayed the real wedged guard.db end to end via the harness in the next PR

## Release notes

- [x] conventional commit title matches semver intent (fix)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #393, opened 2026-07-20
fix(managedstream): drain past wide receipt ranges, restore batch limit after reductions

michiosw:
## Summary

Wires the store-level receipt-range guard (previous PR) into the flush drain and fixes the drain's degenerate batching:

- Pass `ReceiptLimit` (`maxPayloadReceipts`) to `LedgerBatch`; on `ErrLedgerReceiptRangeTooLarge` halve the batch limit, and at the minimum batch advance the cursor past the poison action instead of retrying it forever.
- Restore the batch limit **additively** (+1 per accepted post, AIMD): reductions are per-region, and replaying the real wedged guard.db showed a 73,111-action backlog shipping as **73,111 single-action HTTP posts** because one early reduction pinned the drain at limit 1 — the request pattern behind ENG-546. Multiplicative recovery was tried first and oscillated against the same wide receipt range it had just backed off from, running ~3x slower than no recovery at all.
- Add an env-gated replay harness (`TestReplayLiveDB`) that drains a copy of a real guard.db against a fake ingest server.

## Verification

- [x] `go test ./...`
- [x] Replayed the full 8-day production backlog (73k actions, ~480 MB) from this branch: drains clean end to end, `Flush` returns nil, cursor advances July 12 → July 20, including resume after a mid-drain interruption
- [x] 413-retry and limit-recovery behavior covered by unit tests

## Release notes

- [x] conventional commit title matches semver intent (fix)

### DIFF
--- a/internal/guard/store/sqlite/ledger.go
+++ b/internal/guard/store/sqlite/ledger.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 )
@@ -15,6 +17,7 @@ type LedgerExportOptions struct {
 	UpdatedAfter   *time.Time
 	UpdatedAfterID string
 	Limit          int
+	ReceiptLimit   int
 }
 
 type LedgerBatch struct {
@@ -37,9 +40,12 @@ type LedgerCursor struct {
 const (
 	ledgerCursorUpdatedAtKeyColumn = "__cursor_updated_at_key"
 	ledgerTimestampCursorKeyLayout = "2006-01-02T15:04:05.000000000Z07:00"
+	maxSQLQueryParams              = 500
 )
 
 var ledgerFallbackTimestamp = time.Unix(0, 0).UTC()
+
+var ErrLedgerReceiptRangeTooLarge = errors.New("ledger receipt range exceeds limit")
 
 const authorizationActionsSelect = `
 select id, session_id, turn_id, tool_use_id, trace_id, span_id, parent_span_id,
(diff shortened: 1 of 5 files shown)


michiosw:
* **#394** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/394?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#393** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/393?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/393?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#392** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/392?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

hasandemirkiran:
**Review note — data completeness on the audit ledger (worth a conscious decision)**

At `limit == 1`, when a single action's receipt rowid range still exceeds `maxPayloadReceipts`, `advancePastMinimumBatch` advances the cursor past that action — which permanently skips that action's **own** receipts too, since the cursor never revisits it. The wide count is driven by *foreign* interleaved receipts (those get exported under their own actions' pages), but the poison action's receipts are dropped from the export for good.

That's the right availability call to avoid a repeat of the 8-day wedge, and it mirrors the existing oversized-payload behavior. Flagging only because this is a compliance/audit ledger — silently dropping receipts should be a decision the team explicitly signs off on, and ideally the hosted side can surface the gap (e.g. a receipt-chain discontinuity marker) rather than it being invisible.

Everything else here is solid — the AIMD additive recovery is the right shape, termination is guaranteed (limit strictly halves to 1), and the replay harness is a nice reproduction tool.

hasandemirkiran (approved):

PR #394, opened 2026-07-20
fix(managedobserve): log consecutive stream flush failures unconditionally

michiosw:
## Summary

Non-auth flush errors only reached the env-gated diagnostic logger, so the daemon burned CPU on a wedged export for **8 days** with nothing in the launchd log but malloc noise. Surface consecutive flush failures unconditionally with the same 3rd-then-every-50th cadence already used for auth failures.

## Verification

- [x] `go test ./...`
- [x] Existing managedobserve daemon tests pass

## Release notes

- [x] conventional commit title matches semver intent (fix)

### DIFF
--- a/internal/guard/store/sqlite/ledger.go
+++ b/internal/guard/store/sqlite/ledger.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 )
@@ -15,6 +17,7 @@ type LedgerExportOptions struct {
 	UpdatedAfter   *time.Time
 	UpdatedAfterID string
 	Limit          int
+	ReceiptLimit   int
 }
 
 type LedgerBatch struct {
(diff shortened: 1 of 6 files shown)


michiosw:
* **#394** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/394?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/394?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#393** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/393?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#392** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/392?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

hasandemirkiran:
**Review note — the new log can read as "failing" while it's actually recovering**

Good change — surfacing non-auth flush failures unconditionally is exactly what was missing during the 8-day wedge.

One cross-stack interaction to be aware of (from #393): when the drain *advances past* a poison row, `Flush` returns an error even though it made forward progress that tick. The daemon treats that as a consecutive flush failure, so this new line will emit:

```
managed stream flush failing (%d consecutive): ...
```

while the export is in fact recovering one poison row per interval. Not a correctness bug, but an operator watching the launchd log could misread steady recovery as a stuck export. Might be worth distinguishing "advanced past rejected batch" (progress) from a true no-progress failure so the count only climbs when nothing drained.

hasandemirkiran (approved):

PR #412, opened 2026-07-27
feat(policy): resolve enforcement from remote rollout mode

tumberger:
## Summary

Adds a third managed mode, `remote`: instead of pinning the guard posture at install time, the endpoint follows the rollout mode of the Cedar policy deployment it already fetches (~60s refresh). Flipping the deployment between observe and enforce in the dashboard changes endpoint behavior within one refresh interval — no reinstall, no config push.

- `managedconfig`: accept `"mode": "remote"` (`observe`/`enforce` stay static pins)
- `guard/server`: the boolean Cedar cutover gate becomes an enforcement source (`off`/`static`/`remote`); `remote` claims authority exactly when the fetched deployment (or last-known-good) says enforce
- `guard/runtime`: results carry `mode: enforce` when a Cedar decision was applied under an enforce rollout, so hook edges can recognize authoritative decisions. The static enforce transform now stamps results too — previously enforce-mode daemon results were never stamped, so the hook lifecycle treated them as non-authoritative
- `lifecycle`: in remote mode, only daemon results stamped enforce pass through; everything else is downgraded to observe. When the daemon is unreachable, the hook fails closed only while the cached policy distribution claims enforcement, and fails open otherwise

Fail-state semantics in remote mode: before any deployment is fetched the endpoint observes (a fresh install never hard-blocks); once an enforce deployment is cached, stale/expired cache holds enforcement (deny with `enforcement_not_ready`) rather than silently degrading to observe.

## Verification

- [x] `go test ./...`
- [x] `go test -race ./...`
- [x] `go vet ./...`
- [ ] `buf generate` if protobuf or generated code changed (n/a)

## Release notes

- [x] conventional commit title matches semver intent

### DIFF
--- a/internal/cedarpolicy/mode.go
+++ b/internal/cedarpolicy/mode.go
@@ -1,0 +1,23 @@
+package cedarpolicy
+
+import "github.com/kontext-security/kontext-cli/internal/cedareval"
+
+// DeploymentClaimsEnforce reports whether the cached policy distribution asks
+// this endpoint to enforce: an active (or last-known-good) deployment exists
+// and its rollout mode is enforce. This is the remote-managed counterpart of
+// the static cutover gate, with deliberately different unknown-state
+// semantics: a remote-managed endpoint without a trustworthy deployment stays
+// in observe, because the organization's intent is unknown until a deployment
+// has been fetched. Expiry and staleness do not relinquish enforcement — once
+// an enforce deployment is cached, only an explicit disabled/no-active-policy
+// state or an observe deployment hands authority back.
+func DeploymentClaimsEnforce(snapshot Snapshot) bool {
+	if snapshot.State == StateDisabled || snapshot.State == StateNoActivePolicy {
+		return false
+	}
+	deployment := snapshot.Deployment
+	if deployment == nil {
+		deployment = snapshot.LastKnownGood
+	}
+	return deployment != nil && deployment.RolloutMode == cedareval.RolloutModeEnforce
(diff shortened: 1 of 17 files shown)


tumberger:
* **#413** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/413?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#412** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/412?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/412?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tumberger:
Re the outage-fallback cache path finding: the fallback resolves the cache from `DefaultDBPath()`, which honors the same `KONTEXT_MANAGED_OBSERVE_DB` override the daemon derives its cache path from — the same mechanism that already keeps `SocketPath` aligned between the daemon and hook processes. The programmatic `DaemonOptions` path overrides are test seams and are never paired with a production hook lifecycle. Documented this contract explicitly on `remoteEnforceFromCache` in the latest push: a deployment relocating daemon state must export the env var to hook processes too.

tumberger:
@greptileai review

tumberger:
@greptileai review

tumberger:
@greptileai

PR #428, opened 2026-07-29
feat: compute the risk annotation in the decision path and upload it

hasandemirkiran:
> **Stacked on #423** (which is stacked on #414). Base is `feat/v2-guardrail-llm`; this PR's diff is only the move in-path plus the upload.

## Summary

Moves the classifier from a post-decision observer to an **in-path stage**, running after the policy layer has produced its final answer and before the tool executes, so verdicts reach the hosted ledger attached to the decision instead of trailing it. It stays **strictly advisory** — no gating, no config that lets it deny.

## Advisory by construction, not convention

Annotation happens in exactly one place — `guardHookRuntime.annotate`, between the final decision and `SaveDecision`. It receives an already-settled decision and may write nothing but the `Classifier` field, so there is no code path from a verdict to an allow/deny outcome.

That placement is load-bearing for a second reason. Annotation initially lived inside the local policy provider, injected via a `policy.(RiskPolicyProvider)` assertion that ran *after* Cedar had already wrapped the provider — so on any runtime with a snapshot source the assertion failed and nothing was annotated at all. Every enforce and managed deployment recorded zero verdicts. `TestCedarEnforceDenyIsStillAnnotated` drives a real Cedar enforce deny end to end and reports `verdicts = 0` against that wiring.

`TestAnnotationCannotChangeDecision` runs identical commands through a server with the classifier on and one with it off, and requires byte-identical decision, reason, and reason code.

## Latency

Per gated bash command, end to end (hook process → socket → RuntimeCore → policy → classifier → SQLite), 40 samples after warm-up, local Qwen3-0.6B-Q8:

| configuration | p50 | p95 |
|---|---|---|
| no LLM at all | 21.8 ms | 23.2 ms |
| **classifier on, warm LLM** | **64.2 ms** | **75.6 ms** |
| classifier on, SVM only (LLM shed) | 21.7 ms | 23.7 ms |
| the JSON judge this replaced | ~200 ms | ~247 ms |

**The LLM adds p50 42 ms / p95 52 ms** per command. Even in-path that is under half what the JSON judge cost, because a one-word answer generates far fewer tokens than a JSON object. Denied commands now pay this too (see below); the SVM-only row above is what they used to cost.

## Bounded three ways, because it is now on the hook path

- **Hard 400 ms per-call budget** (~9× warm latency).
- **Circuit breaker** — 3 consecutive failures opens it; one probe admitted after a 30 s cooldown; a failure while half-open reopens immediately.
- **Readiness probe** against `/v1/models` before the first call and after every reopen. This is the one that matters most: a `llama-server` loading its model takes *seconds*, so without it every command during startup would pay the full budget and fail anyway. `TestUnreachableGuardrailCostsNothingAfterBreakerOpens` asserts six commands against a dead endpoint finish well under a second.

Both models run for **every** outcome, denies included — a deny is where agreement between the policy and the models is worth the most, and the decision is already final by the time the classifier sees it. That is a review change from the original design, which skipped the LLM on denies; the cost is the p50 42 ms above now applying to denied commands as well.

## Remote kill switch

A guardrail flag rides the same endpoint-configuration refresh as payload-capture mode, and is persisted so it survives an offline restart. **Absent reads as enabled** — the flag exists to turn the LLM off, so a server that does not carry it must not silently disable the second opinion. An explicitly set `KONTEXT_RISK_CLASSIFIER_MODE` pins the local value and wins, so a developer debugging their own machine is not flipped by a config refresh. The embedded SVM is never gated.

**The remote half is wired but not yet effective, and needs one coordinated server change.** `ConfigIdentity` is a shared hash — the server sends it as the ETag and `Response.Validate` rejects any response whose identity does not match the locally recomputed one — and its preimage covers `payloadCaptureMode` alone. So flipping only `guardrailLlmEnabled` leaves the ETag unchanged and a conditional refresh reuses the cached config, while adding the field to the preimage here alone would break `Validate` against every response the current server sends. Both halves have to move together, with `ResponseVersion` and `identityDomain` bumped. Inert until then, since the server does not send the field; documented at the field and in `docs/guard.md`, and the local override is the working path meanwhile.

## Upload contract

The annotation rides **inside the Decision Fact**, as a `classifier` section on the same `request.decided` fact, with exactly the agreed field names:

```json
"classifier": {
  "svm": {"verdict": "not_risky", "score": 0.0001, "threshold": 0.4, "model_version": "0.1.0"},
  "llm": {"verdict": "risky", "model": "Qwen/Qwen3-0.6B-GGUF", "duration_ms": 44, "cached": false},
  "llm_error": null
}
```

One decision, one copy. It previously sat in its own action column, uploaded as a sibling of the fact and protected by a separate `classifier_hash` in the receipt; that column and that hash are both gone. The receipt already signs the complete fact, so a nested section is tamper-evident without a hash of its own, and there is no longer a way to see one verdict in the fact and a different one beside it. The fact's Go type carries only contract fields, so the local-only halves — the redacted command, its hash, `agent_task`, `llm_prompt_id` — cannot reach the wire by construction rather than by remembering to strip them. A call that was not scored carries `"classifier": null`.

**This half cannot merge alone.** `internal/ledgerfact` mirrors the server-side shared contract and its golden corpus is byte-pinned by `FixtureDigest`, asserted on both sides. The matching hosted change (`packages/shared/src/ledger/decision-fact.ts` + `fixtures/decision-fact-v1.json` + ingest) has to land with it; new digest is `c5904f138bcf2f049ecca7623e83c857e10691762a1e7460b7af7c35136b1797`. Deploy order matters in one direction only: the server must accept the field before CLIs emit it.

`user_feedback`, `feedback_at`, `agent_task`, the redacted command, and `llm_prompt_id` stay local. Note for whoever consumes this: `cached: true` means the verdict was reused from a local LRU for a byte-identical repeat — it says nothing about the model.

## Net simplification

Synchronous classification **removed** machinery: no queue, no worker pool, no drop counter, no drain-on-shutdown, and the local verdict row is written on the same path as the decision instead of racing it. It also removes the shutdown race #423 had to fix, since there is no longer a worker that can outlive the store.

## Verification

- [x] `go test ./...` serially, `-race` on touched packages
- [x] `go vet ./...`, `gofmt` clean
- [x] `guard smoke-test`, `make guard-e2e`
- [x] live measurement against a real `llama-server`
- [ ] `buf generate` — n/a

Two test stubs needed fixing: they counted the readiness probe as an inference. Not a product issue, but worth knowing if you add stubs.


hasandemirkiran:
* **#433** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/433?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#428** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/428?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/428?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#423** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/423?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#424](https://github.com/kontext-security/kontext-cli/pull/424) <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/424?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#414** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/414?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

internal/managedobserve/daemon.go
```diff
@@ -306,6 +309,14 @@ func captureConfiguration(snapshot endpointconfig.Snapshot) payloadcapture.Runti
 	})
 }
 
+// guardrailLLMEnabled reads the org's guardrail kill switch off the same
+// endpoint configuration that carries payload capture. Absent or unconfirmed
+// reads as enabled — see riskclassifier.ResolveLLMEnabled for why that is the
+// opposite of capture's fallback.
+func guardrailLLMEnabled(snapshot endpointconfig.Snapshot) bool {
+	return riskclassifier.ResolveLLMEnabled(snapshot.Config.GuardrailLLMEnabled, snapshot.Confirmed)
```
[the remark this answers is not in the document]
hasandemirkiran: Correct, and a sharper point than the rule I'd written. Fixed in `9bf9f18`: it now reads the **persisted** (`Configured`) directive instead of the effective `Config`, which falls back to defaults while unconfirmed. That satisfies both constraints at once — an explicit `false` survives restarts and unconfirmed fetches, and absence still never disables, so a transient fetch failure can't quietly kill the classifier either. `ResolveLLMEnabled` no longer takes the confirmed flag, since a persisted directive is honoured regardless of it.

srt-kontext-hook.log
```diff
@@ -0,0 +1,294 @@
+hook-start
+sh: /usr/local/bin/kontext: No such file or directory
+exit:127
```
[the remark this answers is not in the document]
hasandemirkiran: Fixed — that was my mistake. A `git add -A` from the worktree root swept the file in; it's removed from the commit and the branch has been force-pushed, so it's gone from the remote.

hasandemirkiran:
@greptileai

internal/endpointconfig/contract.go
```diff
@@ -23,6 +23,15 @@ var sha256HexPattern = regexp.MustCompile(`^[a-f0-9]{64}$`)
 
 type Config struct {
 	PayloadCaptureMode payloadcapture.Mode `json:"payloadCaptureMode"`
+	// GuardrailLLMEnabled is the org's kill switch for the risk classifier's
+	// guardrail LLM. Optional, and absent means enabled — note this is the
+	// OPPOSITE default to PayloadCaptureMode, deliberately. Capture falls back
+	// to the privacy-safe "record nothing" whenever the configuration is
+	// unconfirmed; this flag exists only to turn the LLM off, so treating
+	// "unconfirmed" as off would let a transient fetch failure silently disable
```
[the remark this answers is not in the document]
hasandemirkiran: Correct, and it is worse than a stale gate — thanks, this one is not fixable from this repo alone.

`ConfigIdentity` is a **shared** contract hash, not a local one: the server computes it, sends it as the ETag, and `Response.Validate` **rejects** any response where `ConfigIdentity != ComputeIdentity(r.Config)`. So the two sides must compute it identically.

That means both directions are broken until the contract moves in lockstep:

- Leave the preimage alone (as now) and an org that flips only `guardrailLlmEnabled` gets an unchanged identity → `If-None-Match` matches → 304 → the cached config is reused and the gate never changes. Exactly what you describe.
- Add the field to the preimage here alone and the CLI's recomputed hash stops matching every response the current server sends, so `Validate` fails and endpoint configuration breaks entirely.

The fix has to be one coordinated change: the server includes `guardrailLlmEnabled` in its identity preimage, the CLI matches, `ResponseVersion` (currently 1) and `identityDomain` (`kontext:endpoint-config:v1`) bump, and the portable golden vectors regenerate. That is the same owner who has to add the field to the response in the first place, so it is one ask rather than two.

Flagging to @hasandemirkiran for the call on whether to ship the remote half now with this limitation documented, or hold it until the contract can carry it. The local `KONTEXT_RISK_CLASSIFIER_MODE` override works either way, and the flag is inert until the server sends it.

internal/guard/app/server/policy.go
```diff
@@ -30,19 +33,75 @@ func (p RiskPolicyProvider) DecideHook(ctx context.Context, event risk.HookEvent
...
+// cost only; annotating cannot affect any decision either way.
+func (p RiskPolicyProvider) annotate(ctx context.Context, event risk.HookEvent, decision *risk.RiskDecision) {
+	if p.classifier == nil {
+		return
+	}
+	command := risk.CommandFromInput(event.ToolInput)
+	if strings.TrimSpace(command) == "" {
+		return
+	}
+	allowLLM := decision.Decision != risk.DecisionDeny
```
tumberger: **Requested change:** Is skipping Qwen on Cedar denies intentional? The annotation is useful for denied Bash calls too: Cedar remains the sole authority, while we retain the same SVM + LLM risk evidence for every decision outcome. Please run Qwen for denies as well.

internal/guard/app/server/server.go
```diff
@@ -102,12 +124,20 @@ func NewServerWithPolicyAndOptions(store *sqlite.Store, policy PolicyProvider, o
...
-	guardrail := newGuardrail(opts.RiskClassifier)
-	classifier := newRiskClassifierObserver(store, opts.RiskClassifier, guardrail)
+	// One instance, shared: the runtime feeds it session prompts and the policy
+	// provider reads them back when annotating, so a second instance would
+	// silently lose every agent_task.
+	if opts.RiskClassifier != nil && opts.RiskClassifier.Gate == nil {
+		opts.RiskClassifier.Gate = riskclassifier.NewLLMGate()
+	}
+	classifier := newRiskClassifier(opts.RiskClassifier)
+	if provider, ok := policy.(RiskPolicyProvider); ok && provider.classifier == nil {
```
tumberger: **Suggestion — would you agree?** Move annotation to the shared hook runtime, after Cedar has produced its final result and immediately before `SaveDecision`. With the current injection after Cedar wrapping, managed/enforce paths can bypass the classifier. Runtime placement gives observe and enforce calls one annotation path while keeping the classifier unable to affect Cedar allow/deny.

internal/guard/store/sqlite/store.go
```diff
@@ -1129,6 +1146,11 @@ func receiptInputFromAction(action map[string]any, receiptType string, now time.
 		// than only mirroring selected columns whose meaning may evolve.
 		actionPayload["decision_fact"] = json.RawMessage(decisionFactJSON)
 	}
+	if classifierJSON := stringValue(action["classifier_json"]); classifierJSON != "" {
+		// The advisory verdict is hashed into the same signed payload, so it is
+		// tamper-evident alongside the decision it annotates.
+		actionPayload["classifier_hash"] = hashString(classifierJSON)
```
tumberger: **Requested change:** Please make the classifier part of the signed Decision Fact, then derive any local/queryable mirror from that fact and remove this separate hash. Intuitively, one decision should have one source of truth: we should never be able to see one classifier result in the fact and another in a parallel field. The receipt already protects the complete fact.

internal/guard/app/server/policy.go
hasandemirkiran: It was intentional, but your reasoning is better and it is changed in 7c378a0 — both models now run for every outcome. Same evidence for denies as for allows, and since annotation happens after the decision is final there is nothing for it to affect.

Cost, for the record: a denied bash command now pays the same p50 42 ms / p95 52 ms the allowed path already paid, instead of ~0. `docs/guard.md` records that trade rather than leaving the old "nothing on denies" claim standing.

internal/guard/app/server/server.go
hasandemirkiran: Agreed, and it turned out to be more than a placement question — done in 7c378a0.

The injection was `policy.(RiskPolicyProvider)`, and that ran *after* `newCedarPolicyProvider` had already wrapped the provider. Any runtime with a snapshot source gets a `*cedarPolicyProvider`, so the assertion failed and the classifier was never attached: every Cedar-configured runtime recorded no annotations at all, rather than annotating in the wrong place. `TestCedarEnforceDenyIsStillAnnotated` drives a real Cedar enforce deny end to end and reports `verdicts = 0` against the old wiring.

Annotation now happens in `guardHookRuntime.annotate`, between the final decision and `SaveDecision` — one call site for observe, enforce, and managed alike. `TestAnnotationCannotChangeDecision` still holds.

internal/endpointconfig/contract.go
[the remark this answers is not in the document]
hasandemirkiran: Documented in edb2d13, which was the part I could own here. `docs/guard.md` said the flag "rides the endpoint-configuration sync", which reads as though remote toggling already works; it now states plainly that the remote half is wired but inert, why neither side can move alone, and that `KONTEXT_RISK_CLASSIFIER_MODE=off` is the working path meanwhile. The same note sits on the `GuardrailLLMEnabled` field so it is unmissable from the code.

Leaving this thread open for @hasandemirkiran on the tracking item — filing it against the endpoint-configuration owner is their call, and it is the same ask as adding the field to the response.

internal/guard/store/sqlite/store.go
hasandemirkiran: I agree with the principle, and I do not think this one can ship from this repo — flagging it for @hasandemirkiran rather than half-doing it.

One factual note on the premise first: there is already a single source of truth. `classifier_json` is one column, and both things you can see derive from it — the uploaded action event's `classifier` block is that column under its wire name, and `classifier_hash` is `hashString` over those same bytes. So the fact and the parallel field cannot currently disagree; the objection is structural rather than a live divergence.

What moving it into the fact costs: `internal/ledgerfact` is a mirror of the server-side shared contract, and `testdata/decision-fact-v1.json` is byte-pinned by `FixtureDigest`, which the contract pins to the same value on the other side. Adding a `classifier` section changes the corpus bytes, so CI fails on both sides until both move together — the check is designed to make exactly this impossible unilaterally. It also runs against the shape this PR was asked to emit: an exact `classifier` block on the action event whose names server-side ingest accepts as-is.

So it is the same class as the `ComputeIdentity` thread above: correct, and gated on one coordinated change with the contract owner (fact schema + TS mirror + regenerated corpus + `SchemaVersion` bump), not something to bolt on here. My suggestion is to land the classifier as an action-event annotation now and fold it into the fact when the schema next moves, but the call is @hasandemirkiran's.

hasandemirkiran: Implemented in 3cfca24 — you were right that this is the better shape, and it turned out cleaner than I expected.

The annotation is now a `classifier` section on the Decision Fact. The separate action column and `classifier_hash` are both gone, so there is one copy on the row and the receipt covers it by signing the fact whole. A side benefit I had not anticipated: the fact type carries only contract fields, so the local-only halves (redacted command, its hash, agent task, prompt id) can no longer reach the wire by construction, where before it depended on `json:"-"` tags staying correct. The block's field names are unchanged — only its location moved.

Two fixtures added (a scored call, and one where the LLM was shed) and `FixtureDigest` regenerated to `c5904f138bcf2f049ecca7623e83c857e10691762a1e7460b7af7c35136b1797`. Since the digest is asserted on both sides this cannot merge alone; the matching hosted change is going up as a PR stacked on kontext#751, and the server must accept the field before any CLI emits it.

Leaving the thread open until that one is up so it does not read as done on one side.

internal/endpointconfig/contract.go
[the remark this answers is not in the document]
hasandemirkiran: Tracking item captured, so resolving this. kontext#751 is the coordinated change, under ENG-613: it adds the flag to the endpoint-config response and to the identity preimage, which is exactly the fix described above.

Three corrections went to that PR rather than here, since they are server-side: it must serve the existing shape unchanged under `response_version=1` (the CLI decodes with `DisallowUnknownFields`, so an old binary hard-fails on a new key before any version check runs, and every existing identity hash changes), the flag gates only the guardrail LLM rather than the whole classifier (the SVM is embedded and always runs), and it defaults to on rather than off since this is a kill switch, not an opt-in rollout.

Remaining on this side, once that lands: send `response_version=2` and include the flag in the local preimage. Not done here because there is nothing to match yet. The limitation is documented at the field and in `docs/guard.md` until then, and the local override works throughout.

internal/guard/store/sqlite/store.go
hasandemirkiran: Resolving — the Go half is implemented in 3cfca24 and the hosted half is specced and going up stacked on kontext#751. Recorded in the PR description, including the new corpus digest and the server-first deploy order, so it cannot be lost if this thread closes.

hasandemirkiran:
@greptileai

hasandemirkiran:
@greptileai Re-review please. Two substantive changes since your last pass, plus one item that stays open by decision.

**The risk annotation now lives inside the signed Decision Fact.** It previously sat in its own action column, uploaded alongside the fact and protected by a separate `classifier_hash` in the receipt; that column and that hash are both gone. The receipt already signs the complete fact, so a nested section is tamper-evident without a second hash, and there is no longer any way to see one verdict in the fact and a different one beside it. The fact type carries only contract fields, so the local-only halves of the annotation cannot reach the wire structurally. This half cannot merge alone — the golden corpus digest is asserted in both repos, so the hosted mirror lands with it.

**Annotation moved to the hook runtime, and both models now run for denies.** The earlier wiring injected the classifier via a type assertion that ran after Cedar had wrapped the provider, so every Cedar-configured runtime recorded nothing at all; `TestCedarEnforceDenyIsStillAnnotated` reports zero verdict rows against that wiring.

**The kill-switch identity stays as you describe, deliberately.** `ComputeIdentity` still covers `payloadCaptureMode` alone. Adding the flag here alone would break `Validate` against every response the current server sends, so both sides move together: kontext#751 (ENG-613) is that change, and it has been given three corrections — serve the existing shape unchanged under `response_version=1`, since this CLI decodes with `DisallowUnknownFields` and an old binary hard-fails on a new key before any version check runs; gate only the guardrail LLM rather than the whole classifier, as the SVM is embedded and always runs; and default to on, since this is a kill switch rather than an opt-in rollout. Nothing can go stale today because the server does not send the field yet, the limitation is documented at the field and in `docs/guard.md`, and `KONTEXT_RISK_CLASSIFIER_MODE=off` works throughout. The CLI half — request `response_version=2` and add the flag to the preimage — is a follow-up that must merge after the server half deploys, not part of this stack.

hasandemirkiran:
@greptileai First finding fixed in 7488f86. Second stands by decision, as before.

**Persisted disable re-enabled by a later response** — valid, and reachable now that the server sends the field. The cache remembers the last *explicit* directive across refreshes and restarts, and the daemon resolves that rather than whatever the current response carries. A response that omits the field no longer clears a deliberate disable; the rest of the configuration still tracks the current response. Stored beside the response rather than inside its config, since the response identity is verified against its config on load and editing the config would invalidate it. `TestOmittedDirectiveDoesNotClearAnExplicitDisable` fails on the previous behaviour with `silence cleared the disable`.

The intended consequence, now documented: clearing a disable takes an explicit `true`, not an omitted field. That is the right direction for something whose only job is turning a thing off. Cache file version moves to 2; version 1 files load unchanged with no remembered directive.

**Identity preimage** — unchanged deliberately, and tracked in kontext#751 (ENG-613), which also carries three corrections: serve the existing shape unchanged under `response_version=1`, since this CLI decodes with `DisallowUnknownFields` and an old binary fails on a new key before any version check; gate only the guardrail LLM rather than the whole classifier, as the SVM is embedded and always runs; and default to on, since this is a kill switch. Adding the field to the preimage here alone would break `Validate` against every response the current server sends, so the CLI half — request `response_version=2`, add the flag to the preimage — is a follow-up that merges after the server half deploys.

internal/managedobserve/daemon.go
```diff
@@ -224,6 +226,7 @@ func RunDaemon(ctx context.Context, opts DaemonOptions) error {
 		Interval:       opts.EndpointConfigRefreshInterval,
 		OnChanged: func(snapshot endpointconfig.Snapshot) {
 			host.SetPayloadCaptureConfiguration(captureConfiguration(snapshot))
+			host.SetGuardrailLLMEnabled(guardrailLLMEnabled(snapshot))
 		},
```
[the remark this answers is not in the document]
hasandemirkiran: Fixed — this thread predates the commit. The cache now remembers the last explicit directive across refreshes and restarts, and the daemon resolves that rather than the current response value, so a response that omits the field no longer clears a deliberate disable while everything else still tracks the new response.

It is stored beside the response rather than inside its config, because the response identity is verified against its config on load and editing the config would invalidate it. `TestOmittedDirectiveDoesNotClearAnExplicitDisable` fails on the previous behaviour with `silence cleared the disable`. Intended consequence, now documented: clearing a disable takes an explicit `true` rather than an omitted field.

hasandemirkiran:
@greptileai

hasandemirkiran:
@greptileai Re-review please. One contract change since the last pass: the classifier section on the decision fact now also carries the redacted command and `command_truncated`.

Rationale, since it widens what is uploaded: a verdict without the text it judged cannot be analysed or corrected, and the same command already reaches the ledger as the 240-byte `command_summary` on every action through the same redaction ruleset — this is that text, longer. It is deliberately **not** gated on `payloadCaptureMode`, which governs tool payload records rather than decision evidence. The consequence is stated in the field comment and in `docs/guard.md`: redaction is the only thing between a credential on a command line and the ledger, so `TestUploadedCommandCarriesNoCredential` puts a bearer JWT and an API key through the real save/export path and requires neither to survive.

`command_hash` stays local — it hashes exactly the uploaded text, so it is derivable — as do `agent_task`, `llm_prompt_id`, and the feedback columns. New corpus digest `5efaeebd2519dcdbafaacbe062da1d00fe6923fb3b1bc253a1ad09a3ed8c5e08`; the hosted mirror moves with it.

PR #430, opened 2026-07-30
feat(agent): add a Devin hook adapter

tumberger:
Registers a `devin` agent adapter so `kontext hook --agent devin` can translate
the Devin runtime's plugin hook payloads. Policy evaluation is untouched — the
adapter decodes the incoming event and encodes the decision in the form that
runtime reads from stdout. Same shape as the existing Codex adapter.

## What's here

- `internal/hookruntime/devin.go` — decode/encode
- `internal/agent/devin/` — registration, plus the blank import in `main.go`
- `hook.ErrSkipEvent` — a skip contract honored by both hook runners

## Decisions worth reviewing

**`prompt_id` is dropped for now.** It's a useful per-turn correlator, but it has
no home on `hook.Event`, and folding it into `ToolInput` would be wrong:
`ToolInput` becomes the Cedar evaluation context, so a synthetic attribute there
would be policy-visible and would change the parameters hash recorded on every
decision. A first-class field is a follow-up.

**Only pre-tool-use carries a decision.** Every other event is observation and
emits no directive. Anything other than an explicit allow blocks, so an
unrecognized decision value fails closed.

**An updated tool input is refused, not applied.** That runtime replaces
`tool_input` wholesale rather than merging, so emitting a partial rewrite would
silently drop every field the policy didn't restate, and the post-tool-use hook
would then observe only the rewritten input — leaving the original absent from
the ledger. Nothing populates `UpdatedInput` today, so this is a guard against a
future change rather than active logic. Input rewriting is worth supporting
properly later, with the pre-rewrite input as the authoritative ledger entry.

**`hook.ErrSkipEvent` is new, and touches shared runners.** Some agent runtimes
read a non-zero hook exit as "block the tool call", so reporting an unhandled
lifecycle event as a decode failure would stall the agent on a legitimate
signal. Both runners now exit successfully and emit nothing for a skipped event.
Unknown event names skip for the same reason; malformed input still fails, since
that indicates a bug or a contract change rather than an event we chose not to
handle. The change is additive — it only triggers on that specific sentinel, so
existing adapters are unaffected.

## Follow-ups, deliberately not in this PR

Both change behavior for existing agents and deserve their own review:

1. **No `recover()` on the hook path.** An unrecovered panic exits 2, which some
   runtimes read as "block" — so a nil-map bug becomes an opaque stall rather
   than a visible failure. Worth fixing for every adapter, not just this one.
2. **Runner selection.** `kontext hook` uses the managed-observe lifecycle but
   exits 2 on any decode failure; `--mode` gets the runner that always writes a
   decision and exits 0, but bypasses the lifecycle and dials the Guard socket
   with no autostart. Neither branch is quite right for a hook that must never
   stall the agent.

## Testing

New tests cover field mapping, `prompt_id` exclusion from the Cedar context,
large-integer precision, skip vs. failure classification, the block/allow encode
forms, observation events never emitting a directive, rewrite refusal, and
fail-closed behavior on an unknown decision.

`go build ./...`, `go vet`, and `gofmt` are clean. Full `go test ./...` passes
except two pre-existing load-sensitive timing tests in `internal/guard/judge`
(`llama_server_test.go`), which also fail on `main` under full-suite load and
pass in isolation.


tumberger:
* **#430** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/430?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/430?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #433, opened 2026-07-31
feat(config): negotiate endpoint-config v2 and honour the guardrail switch

hasandemirkiran:
Stacked on #428. Activates the remote guardrail-LLM kill switch, which #428 shipped wired but inert.

## Why it was inert

The flag was not in the identity preimage, and that identity is the ETag. So flipping only the flag left the ETag unchanged, the conditional refresh got a 304, and the runtime kept its stale value. Neither side could fix it alone: adding the field here first would have broken `Validate` against every response the old server sent. kontext#751 added the server half, serving v1 and v2 side by side.

## What this does

Requests `response_version=2` and computes the identity over `["kontext:endpoint-config:v2", 2, payloadCaptureMode, guardrailLlmEnabled]`, matching the server byte for byte.

`testdata/portable/v2/identity-v2.json` is the server vector set copied verbatim and byte-pinned — six vectors covering both flag values across all three capture modes. That is what demonstrates the two implementations agree rather than merely resembling each other; the test recomputes every identity and every preimage.

**Released binaries are unaffected.** v1 keeps serving its original shape and its original hashes (`d2ed1e73…`, `f66d2835…`, `88b90b14…`), which is deliberate on the server side — old CLIs decode with `DisallowUnknownFields` and would fail on a new key before they ever checked the version.

## Two consequences worth knowing

The flag is **required** under v2, so a response omitting it is rejected rather than resolved as enabled. The rollback case that motivated remembering the last explicit directive can no longer reach the cache at all.

The remembered directive still earns its place, for a different path: a cache written under an older response version cannot be revalidated, since its identity was computed over a different preimage. It is discarded rather than failing startup, and the remembered value is what carries a deliberate disable across that upgrade instead of silently re-enabling the LLM.

## Verification

- [x] `go test ./...` serially, `-race` on touched packages
- [x] `go vet ./...`, `gofmt` clean
- [x] portable identity vectors reproduce the server hashes exactly
- [ ] `buf generate` — n/a


hasandemirkiran:
* **#433** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/433?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/433?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#428** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/428?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#423** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/423?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#424](https://github.com/kontext-security/kontext-cli/pull/424) <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/424?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#414** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/414?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

tumberger (approved):

hasandemirkiran:
### Merge activity

* **Jul 31, 2:04 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/433).
* **Jul 31, 2:04 PM UTC**: @hasandemirkiran merged this pull request with [Graphite](https://app.graphite.com/github/pr/kontext-security/kontext-cli/433).

PR #434, opened 2026-07-31
fix(managed): survive a mode the running build does not implement

tumberger:
## Problem

An unrecognized `mode` in `managed.json` fails the config load, and that takes the whole endpoint down rather than degrading it. launchd `KeepAlive` respawns the daemon, it exits 1 on the same config, and it settles into a permanent crash-loop:

```
$ launchctl list security.kontext.managed-observe
  "LastExitStatus" = 256;

$ tail ~/Library/Logs/Kontext/managed-observe.log
Error: load managed config: mode must be "observe" or "enforce"
Error: load managed config: mode must be "observe" or "enforce"
Error: load managed config: mode must be "observe" or "enforce"
```

No daemon means no telemetry *and* no enforcement, and nothing self-heals it:

- the Homebrew updater lives inside the binary that will not boot;
- the binary watchdog only reacts to the old binary being replaced, so it cannot help when the **new** binary is the one refusing to start;
- `kontext doctor` reports `config: ERROR` and never mentions that the daemon is dead.

It needs hands on the machine.

This is reachable any time the installed binary is **older than the install that wrote the config**, because `mode` gains values over time. `remote` is the live example: builds from before it was introduced meet it in configs written by a newer `kontext setup` and die on it. Installing an older build onto a configured Mac is enough to trigger it.

## Fix

Treat an unrecognized mode as the downgrade signal it is, rather than a fatal error: fall back to observe, record the raw value, and say so loudly in both places that can.

```
managed config declares mode "remote", which this build (0.14.1) does not
implement; running in "observe" — the installed binary is older than the
install that wrote this config
```

```
$ kontext doctor
Managed observe:
  config: /Users/…/managed.json (self-serve)
  WARNING: config requests mode "remote", which v0.14.1 does not implement —
  running in "observe"; this binary is older than the install, reinstall to catch up
```

Observe **strictly dominates** the old behavior, so this loosens nothing:

| | crash-loop (before) | observe fallback (after) |
|---|---|---|
| enforcement | none — no process | none |
| telemetry | none | full |
| diagnosable | `config: ERROR` only | names the mode and the cause |

A process that never starts cannot enforce, so falling back to observe cannot enforce *less* than today. It only recovers the visibility that makes the skew findable.

This also matches what the rest of the codebase already does with an unknown mode — `loadManagedMode` (`lifecycle.go`) and `ParseMode`'s `""` case both resolve to observe. `managedconfig.Parse` was the one layer that turned it into a fatal.

## The read path is lenient, so the write paths are now explicitly strict

Two write paths had to be closed, both for the same reason: they treated observe as a value the caller had chosen, when it can now be a value the loader substituted.

**1. `rewriteModeLocked`** validated by round-tripping through `Parse`. With `Parse` lenient that check silently *normalizes* an unsupported mode to observe and writes it through as though the caller had asked for it. New exported `ValidateMode` is the gate for write paths, called before anything is persisted; `TestRewriteModeRejectsUnsupportedMode` asserts the file is left byte-identical on rejection.

**2. `migrateSelfServeModeToRemote`** (thanks @greptile-apps) fired on the fallback. It rewrites a user-scope config whose mode is observe to `remote` — and a downgraded config now *looks* like one, so it would:

- overwrite the operator's declared posture with this binary's guess,
- erase the only on-disk record that the install is newer than the binary,
- and clear `UnsupportedMode` on the reload, silencing **both** the startup log and the doctor warning.

On the self-serve path — every `kontext setup` install — that removed the diagnostic this PR exists to add. The migration now skips any config carrying `UnsupportedMode`; `TestMigrateSelfServeModeToRemote/fallback_observe_from_an_unsupported_mode_is_untouched` covers it and fails without the guard (`returned mode = "remote", want the observe fallback`).

A **missing** mode stays a hard error — that is a malformed config, not a downgrade, and the existing required-field test still covers it.

## Rebased onto main's doctor rework

`DoctorStatus` (#441) landed after this branch was written, so the new warning now participates in it: a downgraded endpoint is **unhealthy but not repairable**. Restarting the daemon cannot teach this binary a mode it does not implement — only reinstalling helps — so `--fix` must not claim it has a repair. This matches the convention on main, where every WARNING in `printStatus` clears `Healthy`.

## Tests

- `TestParseUnknownModeFallsBackToObserve` — unknown mode loads, in observe, with the raw value recorded
- `TestParseFutureModeFallsBackToObserve` — the `remote`-style regression
- `TestParseSupportedModesRecordNoFallback` — all three real modes round-trip with no breadcrumb, so doctor cannot cry wolf
- `TestRewriteModeRejectsUnsupportedMode` — write path stays strict, file untouched
- `TestMigrateSelfServeModeToRemote/fallback observe …` — the migration leaves a downgraded config alone, on disk and in memory
- `TestPrintStatusWarnsOnUnsupportedConfigMode` / `…DoesNotWarnOnSupportedConfigMode` — doctor warns, reports unhealthy and not repairable, and only when it should

`gofmt`, `go vet`, and `go test -race ./...` all clean locally.

### DIFF
--- a/internal/managedconfig/config.go
+++ b/internal/managedconfig/config.go
@@ -91,13 +91,18 @@ func ResolvePath() (string, Scope) {
 }
 
 type Config struct {
-	Version             string      `json:"version"`
-	CloudURL            string      `json:"cloud_url"`
-	Mode                string      `json:"mode"`
-	Agent               string      `json:"agent"`
-	Credentials         Credentials `json:"credentials"`
-	Device              Device      `json:"device,omitempty"`
-	LegacyCoworkEnabled bool        `json:"-"`
+	Version     string      `json:"version"`
+	CloudURL    string      `json:"cloud_url"`
+	Mode        string      `json:"mode"`
+	Agent       string      `json:"agent"`
+	Credentials Credentials `json:"credentials"`
+	Device      Device      `json:"device,omitempty"`
+	// UnsupportedMode carries the raw mode this build did not recognize, after
+	// Mode has already been rewritten to the observe fallback. Empty on every
+	// config whose mode this build understands, so callers can treat a
+	// non-empty value as "the endpoint is running degraded" and say so.
+	UnsupportedMode     string `json:"-"`
+	LegacyCoworkEnabled bool   `json:"-"`
 }
 
 type configFile struct {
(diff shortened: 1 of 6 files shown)


tumberger:
* **#435** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/435?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#434** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/434?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/434?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #435, opened 2026-07-31
feat(doctor): identify builds by source revision, not version string

tumberger:
Stacked on #434.

## Problem

The version string cannot identify a build. It is injected at link time (`-X main.version=…`), so it reports whatever the release pipeline chose to call the binary and nothing about what was compiled. Two binaries can carry the same string and different source; every unlabeled local build is just `dev`.

That makes "is the running daemon the same build as the installed CLI" unanswerable from the tools we ship. Today it takes reading the stamp out of both binaries by hand:

```
$ go version -m "$(which kontext)" | grep vcs
  build vcs.revision=cac15fd669a7e4b0bfbdd78413d25fc0999e3a11
  build vcs.time=2026-07-08T14:08:02Z
```

`doctor` does compare daemon vs installed, but only on the version string — so it is blind exactly where labels collide.

## Fix

Read the VCS stamp the Go toolchain already writes into every binary built from a repository, and use it where identity matters. New `internal/buildinfo` wraps `debug.ReadBuildInfo`; unlike a link-time flag, this stamp is not something a pipeline can set to the wrong thing.

```
$ kontext --version
kontext version 0.16.0 (f7be605a)

$ kontext doctor
Managed observe:
  daemon: running (v0.16.0 cac15fd6, pid 4711)
  WARNING: daemon is running build v0.16.0 cac15fd6 but v0.16.0 f7be605a is installed
  — run 'kontext doctor --fix' to restart it
```

`daemon-status.json` gains `revision` and `modified` fields, written from the daemon's **own** build stamp rather than passed in by the caller — the daemon is the authority on what the daemon is.

## Compatibility

Skew detection prefers revision when both sides have one, and falls back to the version comparison otherwise:

| daemon | installed | verdict |
|---|---|---|
| same revision, different version | | not skewed — source is what matters |
| different revision, **same** version | | **skewed** — the case a version comparison cannot see |
| no revision (predates this field) | any | falls back to version comparison, unchanged |
| any | no stamp | falls back to version comparison, unchanged |

A missing stamp is never a mismatch. Builds made outside a repository or with `-buildvcs=false` legitimately have none, and warning on them would make the check unusable.

## Dirty builds are marked, not assumed equal

Equal revisions only prove equal source for a **clean** build (thanks @greptile-apps). For a modified tree the revision names the commit the build was based on, not what was compiled, so two different dirty builds of one commit are indistinguishable by revision alone. The breadcrumb therefore records `modified`, and such builds render as a single token that cannot be misread as an exact identity:

```
kontext version 0.16.0 (cac15fd6+modified)
  daemon: running (v0.16.0 cac15fd6+modified, pid 4711)
```

They are treated as **unproven rather than skewed** — deliberately quiet:

- nothing in the stamp can distinguish two dirty builds of one commit (the toolchain records no content hash), so a warning would be asserting something it cannot know;
- it would fire on every local build, and would leave `doctor --fix` unable to ever verify a restart — `WaitForDaemonRestart` would spin for its full 10s on any dev build.

Release builds are never modified, so the production path is unaffected. The readout carrying `+modified` is what tells a developer the comparison is approximate.

## `doctor --fix` verification

This narrows a hole: `WaitForDaemonRestart` accepted **any** live daemon whenever the version string was not comparable (`dev`), so a daemon that never restarted was reported as restarted. It now waits for the absence of skew, using the same primitive as the readout, so "came back" and "up to date" cannot disagree.

**Known residual:** on a modified local build the check can still be satisfied without a restart, for the reason above. Making that airtight needs a signal other than the build stamp (the pre-restart PID, or the binary's mtime against the daemon's `started_at`) and is left out of this PR.

## Rebase note

Rebased onto main after #440 (`newRootCmd` extraction, legacy commands retired) and #441 (`DoctorStatus`, `repairTargetAvailable`). Two conflicts, both resolved in favor of main's structure:

- `cmd/kontext/main.go` — `buildinfo.Describe(version)` now sets `Version` inside `newRootCmd`;
- `internal/managedobserve/doctor.go` — the skew warning goes through main's two-branch messaging, so a non-repairable install is still told to "restart it through its managing installation" rather than to run `--fix`.

## Note for reviewers

Go skips VCS stamping in a git **worktree** — `.git` is a file there, not a directory — so a worktree build prints the bare version. That is the designed fallback, not a failure. Verified in a normal clone:

```
$ go build -ldflags "-X main.version=0.16.0" -o /tmp/kontext ./cmd/kontext
$ /tmp/kontext --version
kontext version 0.16.0 (923c1445)
$ go version -m /tmp/kontext | grep vcs.revision
  build vcs.revision=923c144522d14eb1ffabc4ee44719a8c31f9c678
```

## Tests

`internal/buildinfo`: stamped / unstamped / unavailable-buildinfo, `modified` handling, `Describe` and `DescribeRevision` rendering, `Short` on arbitrary input. The `debug.ReadBuildInfo` seam exists because a test binary has no stamp of its own.

`daemonSkew` is a pure function taking the installed revision and tree state as parameters, table-tested across all cases above — including the two that motivated the change (same revision/different version, different revision/same version) and both dirty-tree cases (same revision → unproven, different revision → still skewed).

`DaemonStatus` round-trips the stamp through the file, and a breadcrumb written before these fields existed still reads as "no stamp recorded".

`gofmt`, `go vet`, `go test -race ./...` clean.

### DIFF
--- a/cmd/kontext/main.go
+++ b/cmd/kontext/main.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/kontext-security/kontext-cli/internal/agent"
+	"github.com/kontext-security/kontext-cli/internal/buildinfo"
 	"github.com/kontext-security/kontext-cli/internal/claudemanaged"
 	"github.com/kontext-security/kontext-cli/internal/diagnostic"
 	guardcli "github.com/kontext-security/kontext-cli/internal/guard/cli"
@@ -41,9 +42,14 @@ func main() {
 
 func newRootCmd() *cobra.Command {
 	root := &cobra.Command{
-		Use:     "kontext",
-		Short:   "Kontext CLI — governed agent sessions",
-		Version: version,
+		Use:   "kontext",
+		Short: "Kontext CLI — governed agent sessions",
+		// Reported with the source revision appended, because the version
+		// string on its own cannot answer "which build is this": it is a
+		// link-time label, and release channels that name builds by date give
+		// no way to tell two sources apart. Every other use of `version` stays
+		// the bare string — only what a human reads gains the revision.
+		Version: buildinfo.Describe(version),
 	}
 
 	root.AddCommand(setupCmd())
--- a/internal/buildinfo/buildinfo.go
+++ b/internal/buildinfo/buildinfo.go
@@ -1,0 +1,103 @@
+// Package buildinfo exposes the source revision compiled into this binary.
+//
+// The version string alone cannot identify a build. It is injected at link time
+// (`-X main.version=…`), so it carries whatever the release pipeline chose to
+// call the build and nothing about what was compiled: two binaries can share a
+// version string and differ in source, and one source tree can be published
+// under several strings. Release channels that label builds by date rather than
+// by source make this routine rather than theoretical.
+//
+// The Go toolchain already stamps the real answer into every binary built from
+// a repository — `vcs.revision`, plus `vcs.modified` for an uncommitted tree —
+// and it is not spoofable by a link-time flag. This package reads it so the CLI
+// can report which source it is, and so two processes from the same install
+// (the CLI and the long-lived daemon) can be compared for real.
+package buildinfo
+
+import (
+	"runtime/debug"
+	"strings"
+)
+
+// shortRevisionLength matches the abbreviation git uses in log output, which is
+// what a reader will compare it against.
+const shortRevisionLength = 8
+
+// read is a test seam for debug.ReadBuildInfo, which cannot be made to report a
+// VCS stamp from inside `go test` (the test binary has no stamp of its own).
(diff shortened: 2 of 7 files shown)


tumberger:
* **#435** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/435?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/435?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#434** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/434?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #440, opened 2026-08-01
feat!: retire legacy managed sessions

tumberger:
## Summary
- remove the retired OIDC-managed session workflow: login/logout, start --managed, credential exchange, hosted sidecar, and generated AgentService client
- remove the now-redundant local kontext start wrapper and its update checker; install-token kontext setup plus the managed-observe daemon is the supported path
- preserve setup, managed-observe, local hook/socket transport, and the daemon-owned SVM/LLM guardrail runtime
- remove obsolete RPC generation/CI, dependencies, tests, and documentation

## Retirement evidence
Warehouse data showed 0 legacy CLI sessions in the last 30 days; the last observed session was 2026-06-04. This is an intentional breaking retirement.

## Verification
- go test ./...
- go test -race ./...
- go vet ./...
- go build ./...
- go run ./cmd/kontext --help (no start command)


tumberger:
Added retained command-level coverage in a4df2f2 after removing the legacy managed-session path.\n\nCoverage now keeps the behaviors that remain supported: command surface (including absence of start/login/logout), managed-observe selection, hook validation, Claude managed settings, Cowork classification, and fail-open/fail-closed sidecar behavior.\n\nValidated:\n- go test ./...\n- go test -race ./...\n- go vet ./...\n- go build ./...\n\nManual staging validation: kontext version 0.0.0-staging.20260801.20. doctor --fix restarted the previously running v0.16.0 daemon into the staging build; a follow-up doctor confirmed the staging daemon was running with a recent heartbeat and readable managed config/token. The command surface exposes setup, doctor, guard, and claude, and no longer exposes start, login, or logout.

PR #441, opened 2026-08-02
feat(doctor): report core setup health

tumberger:
## Summary
- make managed-observe diagnostics return explicit configured, healthy, and safely-repairable state; unhealthy configured installations now exit non-zero
- inspect the actual Claude managed settings and validate Codex hooks, their feature flag, and configured binary without mutating user configuration
- keep the fix flag narrow: it only restarts a stale daemon, then rechecks health; document the behavior and current Codex support

## Safety and scope
- doctor is read-only unless the fix flag is passed
- local Guard diagnostics remain separate from hosted setup checks
- JSON/stable machine output and broader automated repairs are intentionally deferred

## Verification
- go test ./...

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,10 +48,12 @@ agents, and starts a background daemon.
 kontext doctor
 ```
 
-Use `doctor` to check hook status, daemon health, and the managed export
-backlog. Re-run `kontext setup` to rotate the token. Run `kontext setup
---uninstall` to remove the self-serve installation. Self-serve setup currently
-supports macOS.
+Use `doctor` to check self-serve hook status, daemon health, and the managed
+export backlog. It exits non-zero when a configured installation is unhealthy.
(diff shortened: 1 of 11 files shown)


tumberger:
Follow-up review update: addressed Graphite's unreachable return and Greptile's empty-heartbeat/unstarted-export false-healthy finding. Also made repair eligibility scope- and prerequisite-aware, require a durable self-serve LaunchAgent, separate self-serve vs organization-managed hook checks, validate configured hook binaries, and preserve the new README coverage-matrix structure. Verified with go test ./....

tumberger:
CI root cause fixed: the new automatic-repair eligibility intentionally returns false outside macOS because launchd is unavailable, but three tests hard-coded repairable=true. The assertions are now platform-aware. I verified the exact CI command: go test -race ./... passes locally.

PR #450, opened 2026-08-05
fix(daemon): answer decision hooks before annotation and persistence

tumberger:
## Problem

The managed daemon settles the policy verdict quickly — Cedar evaluation is in-memory — but the hook response then waits synchronously on (1) classifier annotation, including optional guardrail inference, and (2) the guard.db write. On a long-running install the first write after an idle period pulls cold pages of a large database back from disk for **multiple seconds**, while the hook client's PreToolUse budget is **250 ms** (`preToolUseWait`, `internal/managedobserve/lifecycle.go`).

Every post-idle tool call therefore times out and reads as `managed policy daemon unavailable`:

- **observe**: fails open — silent, invisible, but every decision is recorded late by a daemon the hook already gave up on
- **enforce / remote-enforce**: fails closed — **every tool call on the endpoint is denied**, including the agent's ability to diagnose or revert anything

Measured on a real install (1.4 GB guard.db, pure-Go sqlite): ~5.4 s cold decision, 11 ms warm; a profiler sample during the cold window shows ~87% of the time in `pread` against the database file.

## Fix

Extend the ENG-474 async-ingest pattern to decision-gating hooks. The response is settled the moment the verdict is:

- `decideAndRecord` pre-mints the action ID (`sqlite.NewActionID`), returns the decision immediately, and hands annotation + persistence to a `DeferRecord` executor
- the executor (owned by `runtimehost`) runs jobs on a detached context (`context.WithoutCancel`) — a disconnected hook client can no longer abort the write — and `Host.Close` drains pending jobs **before** the store closes
- `SaveDecision` honors a pre-set `decision.EventID`, so the row lands under the same ID the hook response carried
- opt-in via `runtimehost.Options.AsyncDecisionRecording`; **only the managed daemon enables it**. Wrapper-owned sessions (`kontext start`, guard CLI) keep the synchronous path unchanged

The classifier stays advisory by construction — it now cannot even race the response, which is the intent the code comments already document.

## Not in this PR (follow-ups)

- guard.db retention/pruning — the underlying disease; exported rows currently live locally forever
- recording the hook's *effective* outcome when it times out (today the ledger shows the daemon's late verdict, not what the agent experienced)
- distinguishing "daemon slow" from "daemon absent" in the hook's error string
- `EnsureSessionForEvent` still writes synchronously before the decision; small table, but worth measuring

## Testing

- `TestDeferRecordAnswersWithoutPersisting`: response carries a pre-minted `act_` ID with zero store writes; running the captured job persists under that exact ID
- `TestNilDeferRecordStaysSynchronous`: historical behavior unchanged when the executor is nil
- `TestSaveDecisionHonorsPresetEventID`
- full test suites green: `guard/...`, `runtimehost`, `managedobserve`, `localruntime`

### DIFF
--- a/internal/diagnostic/diagnostic.go
+++ b/internal/diagnostic/diagnostic.go
@@ -34,6 +34,18 @@ func (l Logger) Printf(format string, args ...any) {
 	fmt.Fprint(l.out, Redact(fmt.Sprintf(format, args...)))
 }
 
+// LogAlways writes a redacted line even when verbose output is disabled: an
+// enabled logger keeps its configured sink, a disabled one falls back to
+// stderr. For messages that must reach the process log unconditionally —
+// Printf is debug-gated and silently drops them.
+func LogAlways(log Logger, format string, args ...any) {
+	if log.Enabled() {
+		log.Printf(format, args...)
+		return
+	}
+	fmt.Fprint(os.Stderr, Redact(fmt.Sprintf(format, args...)))
+}
+
 var secretPatterns = []*regexp.Regexp{
 	regexp.MustCompile(`(?i)Bearer\s+[A-Za-z0-9._~+/=-]+`),
 	regexp.MustCompile(`(?i)(access_token|id_token|refresh_token|authorization|cookie)=([^&\s]+)`),
--- a/internal/guard/app/server/defer_record_test.go
+++ b/internal/guard/app/server/defer_record_test.go
@@ -1,0 +1,107 @@
+package server
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/kontext-security/kontext-cli/internal/guard/risk"
+	"github.com/kontext-security/kontext-cli/internal/guard/store/sqlite"
+)
+
+// A deferred-recording server must settle the hook response without touching
+// the store: the response carries a pre-minted action ID, and the persistence
+// job — run later, on the executor's schedule — writes the row under that
+// same ID.
+func TestDeferRecordAnswersWithoutPersisting(t *testing.T) {
+	store, err := sqlite.OpenStore(t.TempDir() + "/guard.db")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer store.Close()
+
+	var jobs []func(context.Context) error
+	server, err := NewServerWithPolicyAndOptions(store, nil, Options{
+		DeferRecord: func(job func(context.Context) error) {
+			jobs = append(jobs, job)
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	event := risk.HookEvent{
+		SessionID:     "s1",
+		HookEventName: "PreToolUse",
+		ToolName:      "Bash",
+		ToolInput:     map[string]any{"command": "echo hi"},
+	}
+	decision, err := server.ProcessHookEvent(context.Background(), event)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if decision.EventID == "" || !strings.HasPrefix(decision.EventID, "act_") {
+		t.Fatalf("EventID = %q, want pre-minted act_ id", decision.EventID)
+	}
+	if len(jobs) != 1 {
+		t.Fatalf("deferred jobs = %d, want 1", len(jobs))
+	}
+
+	summary, err := store.Summary(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+	if summary.Actions != 0 {
+		t.Fatalf("actions before job ran = %d, want 0 (response must not wait on the store)", summary.Actions)
+	}
+
+	if err := jobs[0](context.Background()); err != nil {
+		t.Fatalf("deferred job error = %v", err)
+	}
+
+	events, err := store.Events(context.Background(), "s1")
+	if err != nil {
+		t.Fatal(err)
+	}
+	found := false
+	for _, record := range events {
+		if record.ID == decision.EventID {
+			found = true
+		}
+	}
+	if !found {
+		t.Fatalf("no persisted decision with the response's EventID %q; got %d events", decision.EventID, len(events))
+	}
+}
+
+// Without a DeferRecord executor the historical behavior holds: the response
+// waits on the write and carries the persisted row's ID.
+func TestNilDeferRecordStaysSynchronous(t *testing.T) {
+	store, err := sqlite.OpenStore(t.TempDir() + "/guard.db")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer store.Close()
+	server := newTestServer(t, store)
+
+	event := risk.HookEvent{
+		SessionID:     "s1",
+		HookEventName: "PreToolUse",
+		ToolName:      "Bash",
+		ToolInput:     map[string]any{"command": "echo hi"},
+	}
+	decision, err := server.ProcessHookEvent(context.Background(), event)
+	if err != nil {
+		t.Fatal(err)
+	}
+	summary, err := store.Summary(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+	if summary.Actions != 1 {
+		t.Fatalf("actions = %d, want 1 persisted synchronously", summary.Actions)
+	}
+	if decision.EventID == "" {
+		t.Fatal("EventID empty on synchronous path")
+	}
+}
--- a/internal/guard/app/server/defer_record_test.go
+++ b/internal/guard/app/server/defer_record_test.go
@@ -133,6 +133,79 @@ func TestDeferRecordSkipsNonBlockingHooks(t *testing.T) {
 	}
 }
 
+// A deferred record can land after a SessionEnd already closed its session:
+// the job waits on classifier inference and a possibly cold store while the
+// session winds down in order. Late bookkeeping must not undo that close.
+func TestDeferredRecordDoesNotReopenClosedSession(t *testing.T) {
+	store, err := sqlite.OpenStore(t.TempDir() + "/guard.db")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer store.Close()
+
+	var jobs []func(context.Context) error
+	server, err := NewServerWithPolicyAndOptions(store, nil, Options{
+		DeferRecord: func(job func(context.Context) error) {
+			jobs = append(jobs, job)
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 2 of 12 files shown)


tumberger:
* **#450** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/450?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/450?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tumberger:
This is a meaningful improvement: it removes classifier inference and `SaveDecision` from the managed response path. I do not think it is ready to merge yet, because the remaining synchronization still permits the same 250 ms failure mode and introduces ordering/lifecycle gaps.

1. `PreToolUse` is not yet database-free. `runtimecore.EvaluateHook` still calls `EnsureSessionForEvent`, which performs an UPSERT plus SELECT on the same single-connection SQLite store. Deferred decision jobs can hold that connection, so the next hook can queue behind a cold or contended write and still time out. Please add a real socket-level regression test that blocks a deferred write, then proves a second `PreToolUse` responds within the 250 ms budget.

2. Deferral currently applies to `IngestEvent` as well as `EvaluateHook`. A slow deferred `PreToolUse` annotation can let a later `PostToolUse` persist and export first, reversing the causal receipt order. The base policy decision needs ordered admission ahead of outcomes; classifier annotation should be a separate, linked asynchronous record.

3. The returned action ID is not durable at response time. A crash or failed deferred write loses the proposed/decided rows and signed receipt, and errors are only logged. Please make this an explicit product durability choice: use a bounded durable journal/spool before replying, or expose recording loss and a retry/health mechanism. The current in-memory goroutine is not durable.

4. The executor is unbounded: one goroutine per event can amplify a classifier or SQLite backlog. Use a bounded priority queue/worker model with defined overload behavior. Also do not close the store after a drain timeout while `context.WithoutCancel` jobs may still run; stop admission and join workers before dependency teardown.

5. The direct-server unit test proves the pre-minted EventID, but the daemon socket protocol currently does not serialize `hook.Result.EventID` into `EvaluateResult`. Add the field to the wire conversion and an end-to-end socket test proving the response ID matches the eventual decided row.

Suggested shape: the hook path should be normalize -> cached Cedar decision -> response. A bounded ordered recorder can persist a small base decision envelope first, then materialize SQLite/receipts and advisory classifier evidence asynchronously. Keep decision, annotation, and eventual hook-output/execution observations as distinct correlated facts rather than treating a late daemon write as proof of what the host applied.

tumberger:
One additional merge blocker: the defer branch runs only after `policy.DecideHook`. In non-authoritative Cedar mode, a configured legacy judge can still call `judge.Decide` synchronously before the defer branch. The guardrail classifier disables that judge only when the classifier is enabled and configured; a managed judge with `KONTEXT_RISK_CLASSIFIER_MODE=off` leaves LLM inference on the 250 ms path. Please either move that advisory judge off the decision path too, or add configuration-specific coverage that proves no supported managed configuration performs LLM inference before the hook reply.

tumberger:
Clarification to my earlier wording: `PreToolUse` must evaluate Cedar for every incoming action. The intended fast path is normalize -> evaluate the current in-memory/cached *policy snapshot* against that action -> response. It must not reuse a previous Cedar decision; only policy distribution/parsing may be cached.

tumberger:
Concrete revision path for this PR:

1. Do not defer all of `decideAndRecord`. Add a managed `PreToolUse` fast path that normalizes the event without `EnsureSessionForEvent`, evaluates Cedar for that specific event using the current in-memory policy snapshot, and immediately writes the hook response. Cedar evaluation remains per-hook; only the policy/evaluator state is cached.

2. Replace the per-event goroutine with one bounded recorder owned by `runtimehost`. On the fast-path response, enqueue a pre-minted base-decision envelope first. The recorder must preserve enqueue order for a tool/session, so its `request.proposed` and `request.decided` records are committed before a later PostToolUse outcome. Let `SaveDecision` perform the sole session upsert inside its transaction.

3. Split classifier work from base-decision persistence. A separately bounded classifier worker may run after the base decision is admitted; it emits a linked append-only classifier annotation record. Do not wait for classification before saving the base decision and do not mutate the signed base fact afterward.

4. Define queue durability and overload explicitly. An in-memory queue is a deliberate best-effort audit tradeoff, not durable recording. If durability is required, append a small envelope to a journal/spool before replying; otherwise surface loss/retry/health metrics. Enforce mode needs a defined queue-full/persistence failure posture.

5. Stop recorder admission before shutdown, drain/join workers, and only then close SQLite and the judge. A drain timeout must not leave `WithoutCancel` jobs using closed dependencies.

6. Add `EventID` to `EvaluateResult` wire conversions, then prove in an end-to-end Unix-socket test that the returned ID matches the persisted decided record.

Merge-gate tests: block the SQLite writer and classifier, send consecutive PreToolUse calls through the managed lifecycle/socket, assert each response stays within 250 ms; then verify exactly-once ordered base records, eventual linked annotations, bounded queue behavior, and clean shutdown.

tumberger:
Addressed in 00a0053, plus a rationale for each remaining point.

## Addressed

**1. PreToolUse is now store-free before the response** (comment 1's core point, confirmed). `EnsureSessionForEvent` was still an UPSERT+SELECT on the same single-connection store, so a deferred write holding the connection could push a post-idle hook past the client budget anyway. The session upsert now rides inside the same deferred job, ahead of `SaveDecision` — preserving the synchronous order and the semantics `SaveDecision`'s own inline upsert lacks (mode stamp, reopen of a closed observed session). Only the pure session-ID normalization stays on the response path, so the decision, classifier state, and the eventual rows key alike.

On the requested contention regression test: `TestDeferRecordAnswersWithoutPersisting` now asserts **zero session rows and zero action rows** exist before the deferred job runs. With no store access at all on the pre-response path, there is no connection for a deferred write to contend on — the no-store-touch invariant subsumes the timing assertion, without a wall-clock test that would flake in CI. A socket-level end-to-end timing test through the managed lifecycle needs launchd/probe scaffolding and is follow-up material.

**2. Deferral is gated to decision-gating hooks** (comment 2's premise, confirmed). `IngestEvent` did hit the defer branch, which for non-blocking hooks nested a second goroutine inside the async-ingest one that already answered the client — no latency win, and it did reorder rows. Non-blocking hooks now keep their historical synchronous writes (`TestDeferRecordSkipsNonBlockingHooks`).

**3. Lost writes are countable.** A failed deferred job now logs with a running failure total since daemon start, so silent audit gaps are visible in the daemon log rather than one-off lines.

## Deliberately not in this PR

- **Ordered/bounded recorder with per-session sequencing** (revision path #2): the store has no FK constraints between action rows; rows correlate by `tool_use_id`; the receipt hash chain was already interleaved across unrelated sessions before this PR; and the hosted ingest treats arrival order as insignificant by design (idempotent, order-independent upserts). With deferral now limited to blocking hooks, non-blocking rows keep their pre-PR ordering. Sequencing machinery would guard a property nothing consumes. A bounded worker pool is still worth doing as a follow-up — but it mirrors the async-ingest executor already in production, and the single SQLite connection, not goroutine count, is the serialization point.
- **Splitting classifier annotation into a linked append-only record** (revision path #3): declined — this fights the current design. The annotation rides inside the action row and its receipt, giving one atomic signed record; splitting would double receipt volume for an immutability property nothing relies on. The classifier is already advisory by construction (it runs strictly after the decision settles).
- **Durability journal/spool** (revision path #4): losing an audit row on daemon crash is a bounded, now-observable tradeoff, strictly better than the incident this PR fixes (fail-closed on every tool call). A durable spool and the enforce-mode queue-full posture deserve their own design pass, not a bolt-on to an incident fix. Follow-up issue.
- **Shutdown admission gate** (revision path #5): the daemon closes the host with a background context, so the drain always completes before the store closes; the listener is already down before the drain, so no new work can be admitted. The drain-timeout scenario requires a caller that doesn't exist today. Defense-in-depth follow-up at most.
- **`EventID` on the wire + socket E2E test** (revision path #6): the field was never serialized over the socket — this pre-dates the PR — and no client-side consumer reads it (the hook output format has no event-ID field). Good hygiene; follow-up together with the socket test.
- **Judge on the sync path** (the additional merge blocker): the judge is opt-in via explicit env vars, never enabled by default in the managed daemon, advisory by construction (its decision path always returns allow), and its synchronous placement pre-dates this PR. The described combination (classifier off + judge configured) is an explicit operator override, and the judge is slated for retirement. Tracked separately; not a blocker for this fix.

tumberger:
Second hardening pass in 5f28234, after another round of review:

- **Session-reopen race — real, now fixed.** Moving the session upsert into the deferred job (00a0053) introduced a race the original PR didn't have: a deferred PreToolUse write landing after a SessionEnd would reset `status`/`closed_at` and reopen the closed session. The deferred job now uses `BackfillObservedSessionWithMode`, which refreshes agent/cwd/mode but never resurrects liveness — an in-order hook still reopens, late bookkeeping never does. Regression test: `TestDeferredRecordDoesNotReopenClosedSession`. With this, the remaining out-of-order persistence is row *arrival* order only, which every consumer of these rows treats as insignificant by design.
- **Failure logging is no longer debug-gated.** `Logger.Printf` only writes under `KONTEXT_DEBUG=1`, so the failure counter added in 00a0053 was invisible on a production daemon — the "observable" claim was overstated as written. Deferred-record failures now go through a shared `diagnostic.LogAlways` (stderr fallback, redacted), which lands in the daemon log unconditionally. Still process-local and lost on crash; durable loss accounting remains follow-up work with the spool/journal decision.
- **Deferred work is bounded.** A semaphore caps concurrent deferred jobs at 4 — enough to hide store latency, small enough that a hook burst can't stampede the classifier's local model with concurrent inference. Goroutines still spawn per event (they're the cheap part and preserve the non-blocking response), but classifier + write concurrency is now fixed.
- **`EventID` is on the wire.** `EvaluateResult` now serializes `event_id` and both conversions carry it, so the PR's claim that the hook response carries the pre-minted action ID is true end-to-end for the managed daemon, not just for the in-process server. Socket round-trip test: `TestServiceCarriesEventIDOverSocket`.

Unchanged positions: no ordered-recorder machinery (no consumer of these rows depends on arrival order, and the reopen fix removes the one stateful consequence late writes had), and the legacy judge's synchronous placement in `DecideHook` stays a separate work item — it is opt-in via explicit env configuration, advisory by construction, and predates this PR.

PR #454, opened 2026-08-06
fix(endpointconfig): tolerate weak ETags and log refresh failures

tumberger:
## Summary

Client-side half of the payload-capture outage fix (server-side: kontext#799 + the Cloudflare cache rule).

**Incident:** since Aug 5 ~12:56 UTC, edge compression on `GET /installations/:id/configuration` weakened the response ETag to `W/"<identity>"`. `parseETag` rejected that form, every refresh failed with `ETag does not match config identity`, the endpoint-config cache never confirmed, and `SafeRuntimeConfiguration` pinned payload capture to `summary` **for every agent on the endpoint** — with zero log lines from the refresh loop.

## Changes

- **`parseETag` accepts weak ETags** (`W/"…"`): an intermediary that transforms the representation may legally weaken the validator (RFC 9110 §8.8.1); the identity inside is still the origin's. This matches the tolerance `cedarpolicy`'s client already has — the two clients were inconsistent.
- **Refresh failures are now visible:** the `endpointconfig` and `cedarpolicy` refreshers log through the daemon diagnostic — once per **distinct** error plus one recovery line (not once per minute-retry), via `diagnostic.LogAlways` so it reaches the daemon log even without `KONTEXT_DEBUG`.
- Regression tests: weak-ETag fresh + not-modified fetches, `parseETag` form table, failure/recovery logging contract (dedup, recovery, re-log on new error).

## Why client-side too, when the server is being fixed

Kontext runs on customer machines behind corporate proxies and TLS-inspection middleboxes we don't control. Any RFC-compliant intermediary may weaken validators; hard-failing on that is the actual defect. The server-side `no-transform` fix removes the current trigger; this removes the class.

## Testing

`go build`, `go vet`, full `go test ./...` green. Root cause chain verified live against `api.kontext.security` (strong ETag via plain curl, `W/"…"` + `content-encoding: gzip` via `--compressed`, ledger rows carrying `"fallback_reason":"refresh_failed"`).

## Follow-up (not in this PR)

`kontext doctor` currently reports a healthy daemon while capture is degraded; a payload-capture/config-freshness status line would have surfaced this in seconds.

### DIFF
--- a/internal/cedarpolicy/cache.go
+++ b/internal/cedarpolicy/cache.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/kontext-security/kontext-cli/internal/diagnostic"
 )
 
 const (
@@ -331,6 +333,13 @@ type Refresher struct {
 	Interval       time.Duration
 	MaxBackoff     time.Duration
 	Now            func() time.Time
+	// Diagnostic surfaces refresh state transitions in the daemon log. A
+	// refresh loop that fails silently leaves enforcement running on a stale
+	// policy with no operator-visible signal, so failures must not be
+	// log-invisible.
+	Diagnostic diagnostic.Logger
+
+	lastFailure string
 }
 
 func (r *Refresher) Refresh(ctx context.Context) error {
@@ -360,6 +369,7 @@ func (r *Refresher) Refresh(ctx context.Context) error {
 		r.recordFailure(err)
 		return err
 	}
+	r.recovered()
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


tumberger:
* **#454** <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/454?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext-cli/454?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #457, opened 2026-08-07
fix(hook): label Cowork sessions that run on the host

hasandemirkiran:
## Problem

Cowork activity never appears under Cowork in the dashboard. On this machine the ledger holds 22 sessions and **zero** `cowork` rows, ever.

The cause is labeling, not lost data. Cowork opened on a folder or repository runs the bundled CLI **on the host**, so its hooks fire normally and the daemon records everything — but `isCoworkPath` only recognizes the retired layout `~/Library/Application Support/Claude/local-agent-mode-sessions/<acct>/<org>/local_*`. A modern host session looks like this instead:

- `cwd` → the user's own checkout, e.g. `…/kontext-cli/.claude/worktrees/<name>`
- `transcript_path` → `~/.claude/projects/<slug>/<uuid>.jsonl`

Neither matches, so `isCoworkHookContext` falls through and the session is stored as `claude_code`. Filter the dashboard for Cowork and you see nothing, which reads as a broken integration.

Observed live with desktop app 1.24012.9 / claude-code 2.1.219: Cowork session `local_9826bfca-…` landed in the ledger as `efb31c8c-…` with `agent = claude_code` and its Bash/Read actions attached.

## Fix

Identify host-run Cowork by environment rather than by path. The desktop app sets `CLAUDE_CODE_HOST_SESSION_ID` to the Cowork session id, which carries a `local_` prefix; hooks inherit it as child processes, and plain Claude Code leaves it unset. The existing path heuristic stays as a fallback so the older layout keeps working.

## Testing

- New table test covers the real host-session payload shape plus the unset, empty, and non-`local_` cases.
- Confirmed the new test fails on `main` (`agent = "claude"`) and passes here.
- Existing path-heuristic test still passes, pinned against a stubbed environment.
- `go build ./...`, `go test ./...`, `go vet ./...`, `gofmt -l` all clean.

Not covered: the full managed-observe path end-to-end. It has no state-directory override, so exercising it would write into the live ledger and stream to the real dashboard. The unit tests target `DecodeHookInput` directly, which is the function that was wrong.

## Scope

This fixes host-run Cowork only. Cowork sessions that execute in the per-session VM still bypass observation entirely — no kontext binary, managed-settings drop-in, or daemon socket exists in the guest, so no hook can fire. That is a separate gap and is untouched here.

### DIFF
--- a/cmd/kontext/main.go
+++ b/cmd/kontext/main.go
@@ -32,6 +32,8 @@ var version = "dev"
 
 var userHomeDir = os.UserHomeDir
 
+var lookupEnv = os.Getenv
+
 func main() {
 	root := newRootCmd()
 	if err := root.Execute(); err != nil {
@@ -291,6 +293,9 @@ func (a managedHookAgent) DecodeHookInput(input []byte) (hook.Event, error) {
 }
 
 func isCoworkHookContext(input []byte, event hook.Event) bool {
+	if isCoworkHostSession() {
+		return true
+	}
 	if isCoworkPath(event.CWD) {
 		return true
 	}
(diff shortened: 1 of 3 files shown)


hasandemirkiran:
Closing — opened on a mistaken premise.

I assumed the desktop session I was running in was Cowork and that host Cowork was being mislabeled `claude_code`. It wasn't. Real Cowork sessions have cwd under `~/Library/Application Support/Claude/local-agent-mode-sessions/<acct>/<org>/local_*`, which the existing `isCoworkPath` heuristic already matches. Verified against a live session today: `3b487384-892f-4250-a1c9-2c1f6f0e54e2` → `agent = claude_cowork`, tool `mcp__workspace__bash`. Detection works as designed.

Worse than redundant, this would have been a regression. `CLAUDE_CODE_HOST_SESSION_ID=local_*` is set by the desktop app's LocalSessions layer for **both** Cowork and Claude Code sessions, so keying off the `local_` prefix would have relabeled every desktop Claude Code session as `cowork`.

The real gap is elsewhere and needs no change here: Cowork sessions whose execution runs in the cloud (or is proxied to the local VM via `remote-bash`) never invoke a local hook, so there is nothing to label. That is tracked separately.

PR #458, opened 2026-08-10
fix(setup): let `kontext setup` re-run against the active profile

hasandemirkiran:
`kontext setup` refuses to re-run on any Mac that has profiles, including the
plain production case it was meant to serve:

```
workspace Acme (org_...) is already set up as profile "prod" on https://api.kontext.security

Switch to it with `kontext profile use prod`, or remove it first to re-create it.
```

Rotating an install token is the documented reason to re-run setup, so this
blocks the supported path. There is no way around it from the command line —
the guard fires before the target profile is resolved.

## Cause

The duplicate-workspace guard (added in #446) excluded `opts.Profile` — the
profile the *caller named*. A plain `kontext setup` leaves that empty and
resolves its target to the **active** profile, so the scan skipped nothing,
found the very profile the run was about to write, and reported it as a
duplicate of itself.

## Fix

Exclude the profile the run will actually write to, not the one the caller
happened to name. That is knowable before the workspace is in exactly the cases
that need it:

| Call | Writes to | Excluded |
| --- | --- | --- |
| `setup` / `profile add <name>` | that name | that name |
| `setup` (no name) | the active profile | the active profile |
| `profile add` (derived name) | a new profile | nothing |

The derived-name case is why this is not simply "always exclude the active
profile": that run creates a new profile, so every existing one stays a real
candidate.

## Tests

- `TestSetupRerunWithoutAProfileNameRotatesTheActiveProfile` — the regression.
  Verified it fails without the fix, with the refusal quoted above.
- `TestSetupStillRefusesADuplicateWhenTheNameIsDerived` — the guard is not
  blunted: a derived name is still refused even when an active profile exists
  that a plain setup would have excluded. Verified it passes under the revert.

Full suite green.

### DIFF
--- a/internal/setup/inventory.go
+++ b/internal/setup/inventory.go
@@ -238,36 +238,10 @@ func orDash(value string) string {
 //
 // `exclude` is the profile being written, so re-running setup for an existing
 // profile — rotating its token — is never mistaken for a duplicate of itself.
-// rewriteTarget names the profile this run will write to, when that is knowable
-// before the workspace is. It is what the duplicate check must EXCLUDE:
-// rewriting a profile in place is the documented way to rotate its token, and
-// counting the profile you are about to write as a duplicate of itself refuses
-// the re-run that `kontext setup` promises is safe.
-//
-// Empty means "exclude nothing", which is correct in the two cases it returns
-// it for: a derived name (`profile add` with no name) creates a NEW profile, so
-// every existing one is a real candidate; and a legacy install with no active
-// profile is not a profile, so nothing can collide with it.
-func rewriteTarget(opts Options) (string, error) {
-	if opts.Profile != "" {
-		return opts.Profile, nil
-	}
-	if opts.DeriveProfileName {
-		return "", nil
-	}
-	// Plain `kontext setup`: resolveTarget falls back to the active profile, so
-	// that is the profile being rewritten.
-	active, err := profile.ActiveName()
-	switch {
-	case err == nil:
-		return active, nil
-	case errors.Is(err, profile.ErrNoActive):
-		return "", nil
-	default:
-		return "", err
-	}
-}
-
+// profileBoundToWorkspace names an existing profile already bound to this
+// workspace on this backend, or "" if none is. exclude is the profile the
+// caller is about to write to — rewriting a profile in place is how a token is
+// rotated, so it must never count as a duplicate of itself.
 func profileBoundToWorkspace(organizationID, cloudURL, exclude string) (string, error) {
 	if strings.TrimSpace(organizationID) == "" {
 		// Nothing to compare. The legacy env-fallback org reports an empty id, and
--- a/internal/setup/inventory.go
+++ b/internal/setup/inventory.go
@@ -238,6 +238,36 @@ func orDash(value string) string {
 //
 // `exclude` is the profile being written, so re-running setup for an existing
 // profile — rotating its token — is never mistaken for a duplicate of itself.
+// rewriteTarget names the profile this run will write to, when that is knowable
+// before the workspace is. It is what the duplicate check must EXCLUDE:
+// rewriting a profile in place is the documented way to rotate its token, and
+// counting the profile you are about to write as a duplicate of itself refuses
+// the re-run that `kontext setup` promises is safe.
+//
+// Empty means "exclude nothing", which is correct in the two cases it returns
+// it for: a derived name (`profile add` with no name) creates a NEW profile, so
+// every existing one is a real candidate; and a legacy install with no active
+// profile is not a profile, so nothing can collide with it.
+func rewriteTarget(opts Options) (string, error) {
+	if opts.Profile != "" {
+		return opts.Profile, nil
+	}
+	if opts.DeriveProfileName {
+		return "", nil
+	}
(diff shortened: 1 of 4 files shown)


hasandemirkiran:
Good catch — fixed in 8b2ca2c, and the fix is simpler than what it replaces.

The `rewriteTarget` helper read the active profile to pick the exclusion, then
`resolveTarget` read it again to pick the write target. Two reads, so a
`kontext profile use` landing between them sent the write to a profile the
guard never cleared — a workspace bound twice, which is exactly what the guard
exists to prevent. Worth taking seriously because the menu bar app switches
profiles on one click, so the window is reachable rather than theoretical.

Rather than resolve once and thread the value through, the target is now
resolved **once, up front**, and the exclusion is `slot.Profile` — the
exclusion is the target by construction, instead of two code paths having to
agree. `rewriteTarget` is deleted.

Resolving before the guard writes nothing: a `target` is path arithmetic
(`filepath.Join` over `profile.Dir`), so a refused run still leaves the profile
untouched — the property the original ordering was protecting.

The derived-name case is unchanged and still resolved after the ping: its name
comes from the workspace, so it cannot be known earlier, and it excludes
nothing because that run creates a new profile.

Pinned by a test rather than left to construction: `boundProfileLookup` is a
seam at the guard, and `TestSetupWritesToTheProfileItCleared` moves the active
pointer at exactly that instant. Against the previous commit it fails with:

```
setup wrote to profile "other", but cleared "work" past the guard
```

Full suite green.

hasandemirkiran:
@greptileai

hasandemirkiran:
On the follow-up (retained target vs. concurrent rename/removal): the
observation is accurate, but it is pre-existing and not something this PR
introduces, so I don't think it should block.

`main` already resolves the target once and retains it across every write that
follows — keychain item, config, installation identity, workspace cache:

```go
slot, err := resolveTarget(profileName)   // main, ~L250
...
writeKeychainToken(ctx, slot.KeychainItem, token)   // L258
profile.Create(slot.Profile)                        // L281
identityPath := slot.IdentityPath                   // L293
writeWorkspace(slot.Profile, workspace)             // L308
path := slot.ConfigPath                             // L641
```

This PR moves that resolution earlier by the width of one profile scan. It
widens the window marginally; it does not create it.

What is already covered:

- **Removal** — `profile.Remove` refuses the active profile, which is exactly
  the target a plain `kontext setup` resolves. So the reported case cannot
  happen for the command this PR fixes.
- **Active-pointer movement** — the subject of the first finding, now closed
  and pinned by a test.

What is left is `profile rename` of the active profile during an interactive
`setup`. Real, but it is one instance of a general property: no `kontext`
command takes a lock against another, so any pair of concurrent
profile-mutating commands can interleave. Closing it properly means a lock file
across the profile-mutating commands (`setup`, `profile add|use|rm|rename`),
which is a design change with its own failure modes (stale locks, an
interactive `setup` holding one across a token prompt and a `sudo` prompt) —
not something to bolt onto a bug fix whose whole purpose is to unblock a
command that currently cannot run at all.

Happy to open a separate issue for the locking work if you'd like it tracked.

hasandemirkiran:
Fixed in 4c1f3d1 — you were right that it was worth closing, even though it
predates this PR.

The target is snapshotted when it is resolved (the active profile it came from,
and whether it already existed) and confirmed at the last point where nothing
has been written. A `profile rename` or `profile rm` landing in between is
refused:

```
the active profile changed from "work" to "other" while setup was running, so nothing was written; re-run the command
profile "work" was renamed or removed while setup was running, so nothing was written; re-run the command
```

Optimistic rather than locked, deliberately: an interactive `setup` holds the
terminal across a token prompt and a `sudo` prompt, so a lock would block the
menu bar app's `profile use` for as long as someone leaves a prompt sitting
there. This does not prevent the interleaving — it refuses to write through it,
which turns a stray profile nobody notices into an error that says what
happened and can be retried.

Existence is only checked when the profile existed at resolution, so a run that
is *creating* a profile is unaffected.

Two tests, each mutating state through the guard seam:
`TestSetupRefusesWhenTheTargetIsRenamedMidRun` (renames after the lookup, so
the run reaches the write it must refuse) and
`TestSetupRefusesWhenTheTargetIsRemovedMidRun` (removal is only reachable for a
non-active target, since `profile.Remove` already refuses the active one).

`TestSetupWritesToTheProfileItCleared` now asserts the stronger outcome — that
run is refused outright rather than merely landing on the right profile.

Full suite green.

hasandemirkiran:
@greptileai

hasandemirkiran:
That residual is inherent to an optimistic check — any check has a gap between
checking and acting, and only a lock closes it.

What changed is the size of the gap. It was the whole run, spanning an
interactive token prompt and a `sudo` prompt: seconds to minutes, and reachable
by a single click in the menu bar app. It is now the microseconds between the
confirm and the first write.

Closing it entirely means a lock across `setup` and `profile add|use|rm|rename`.
I don't think that trade is right here: an interactive `setup` would hold the
lock across both prompts, so the menu bar app's `profile use` would block for as
long as someone leaves a prompt sitting there — turning a rare, recoverable
interleaving into a routine hang.

Leaving it here. Happy to open a separate issue for the locking work if it
should be tracked.

PR #474, opened 2026-08-20
feat(setup): never overwrite an existing workspace on a plain re-run

hasandemirkiran:
## Problem

Plain `kontext setup --token …` writes the **active** profile, which is only right while the token belongs to the workspace that profile is already bound to. Pasting a token for any other workspace silently rebound the active profile in place: keychain token overwritten, `cloud_url` rewritten to setup's default (production), workspace label replaced — a working workspace destroyed, with exit 0 and green checkmarks. With local + staging + prod profiles now a normal spread, that made plain setup a footgun: an active `local` profile plus a production token quietly became a production profile named `local`.

## Behavior

Plain setup (no name, no derivation) now triages on what the validated token resolves to:

| Token's workspace | Before | Now |
|---|---|---|
| The active profile's own | rotate in place | rotate in place (unchanged) |
| Bound to another profile | error: "already set up as profile X" | rotate **that** profile's token and switch to it |
| New workspace / different backend | rebind the active profile in place | leave it untouched; add a derived-name profile (as `profile add` would) and switch to it |

The refusal in the middle row only made sense for a *named* run, where the name given and the profile found disagree — there it stays. For a plain run the token names the intent completely: connect this Mac to that workspace.

The switch moves the active pointer just before the LaunchAgent reinstall, so the daemon comes up already serving the retargeted profile — one restart.

## Deliberately unchanged

- **First-time setup**: no active pointer still resolves to the legacy slot, exactly as before.
- **`profile add`** (named or derived): duplicate still refuses; re-running for the same profile still rotates.
- **Legacy unprofiled machines**: no recorded binding to compare, so they keep rewriting in place — pinned by a test.
- **Broken profiles**: unreadable config or workspace cache reads as rotate-in-place, keeping setup as the repair path.

## Testing

Four new tests in `recovery_test.go`: switch-on-duplicate (token rotated, previously-active profile untouched, no third profile), add-on-new-workspace (old binding and token survive), cross-backend protection (active local profile keeps its `cloud_url` under a production token), and the legacy in-place pin. Existing concurrency guards (`TestSetupWritesToTheProfileItCleared`, mid-run rename/remove) still pass unchanged. Full `go test ./...`: 37 packages ok.


hasandemirkiran:
@greptileai review

hasandemirkiran:
@greptileai review

hasandemirkiran:
@greptileai review

hasandemirkiran:
@greptileai review

hasandemirkiran:
@greptileai review

Before re-reviewing, please weigh the scope of this PR against the repository's documented concurrency model. The three remaining findings describe interleavings that are pre-existing properties of that model, not defects introduced here:

1. **Rename/removal invalidating a target while the pointer is written** — `profile rename` and `profile rm` have never participated in any lock. The model is explicitly optimistic; see the `targetSnapshot` comment in `internal/setup/target.go`: *"There is no lock between kontext invocations… This is optimistic instead: it does not prevent the interleaving, it refuses to write through it."* `SetActive` re-validates existence under the new pointer lock, which is strictly stronger than both the pre-PR behavior and the long-shipped `profile use` path (`setup.Activate`). A rename can invalidate the active pointer at any moment with or without this PR.

2. **A live claim renamed away and recreated under its old name** — renaming a profile that a concurrent setup is mid-creating is the same operator-level interleaving as (1). The ownership marker guarantees cleanup touches nothing this run does not own; the alleged failure requires a `profile rename` of a half-created profile plus a re-creation at the old pathname inside one failing run's window. The pre-PR code had no protection anywhere on this path.

3. **Rollback racing another selector's agent restart** — the launch-agent lifecycle (`bootout`/`bootstrap`) has never been serialized between kontext processes: two concurrent `profile use` invocations race identically today (`switch.go` takes no cross-process lock). This PR's rollback restarts the agent only after winning the pointer CAS — strictly stronger than the `Activate` path it mirrors.

Scope of this PR: plain `kontext setup` must never overwrite an existing workspace binding. Within that scope it now provides atomicity — flock'd pointer compare-and-set, mkdir-claimed targets, ownership-checked cleanup — exceeding every neighboring path in the package. Serializing the global profile lifecycle (rename/rm/agent restarts under a common transaction) is a separable, repo-wide change that would be wrong to smuggle into this diff; findings that require it apply equally to code on main.

hasandemirkiran:
@greptileai review

PR #510, opened 2026-09-17
fix(authority): preserve hook access and resume reporting after switches

michiosw:
## Where We Are

Managed hooks bypassed the function that measured Full Disk Access, so their ledger metadata omitted it. Resource reports also stayed empty after an organization re-enable, while local opt-out left old server reports visible.

## Where We Want To Go

Record the invoking app's Full Disk Access, resend unchanged reports after re-enabling, and clear this Mac's server report while local scanning is off.

## How do we get there

Probe Mail with open plus one directory-name read in the managed hook process, retaining unknown errors as unknown. Preserve the result through the managed Unix socket and deferred SQLite recording. Policy recovery queues a stream-state reset, avoiding concurrent writers. Every locally disabled heartbeat sends optional `device.authority_scan: false`; an accepted clear resets the last report so enabling sends it again. The paired cloud close-out PR owns the shared contract and server clear.

Validation: `go test ./... -race` and `go vet ./...` pass, including managed hook-to-ledger false metadata, organization recovery within the hourly resend window, repeated local-off markers, and local re-enable. Michel also ran the isolated probe from a shell without Full Disk Access: the real probe and managed socket persisted `full_disk_access: false` beside `permission_mode: default` in the temporary guard.db (1.37 s); no live daemon, profile, config or installed binary is changed. No merge or release.

PR #511, opened 2026-09-18
fix(cowork): evaluate Cedar policies for Cowork hooks

hasandemirkiran:
## Summary

Cowork hooks were omitted from the Cedar v2 agent mapping, so even a harmless `echo hello` or `Read` failed request conversion and was denied whenever the deployment enforced policy. Map `cowork` and `claude-cowork` to the existing Claude Code policy identity while preserving Cowork attribution in hook and ledger events.

Normalize Cowork's `mcp__workspace__bash` to the shell policy resource so shell block rules still apply after conversion succeeds. This alias is limited to Cowork; other agents retain the named MCP resource. Document the policy identity and cover both rollout modes, harmless calls, agent-scoped denies, and force-push enforcement. Existing fail-closed behavior is unchanged.

## Verification

- Regression cases fail before the fix and pass afterward.
- [x] `go test -race ./...`
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `go mod verify`
- [x] `go run ./cmd/kontext guard smoke-test`
- [x] `make guard-e2e`

## Release notes

- [x] Conventional `fix` title matches patch-release intent.

PR #512, opened 2026-09-18
fix(authority): send the local opt-out marker only to servers that know the scan flag

michiosw:
## Mechanism
While `KONTEXT_AUTHORITY_SCAN=off` is set, every heartbeat carries `device.authority_scan: false` so the server clears that Mac's stored report (#510).

## Problem
A server from before the authority scan, which is production today, validates the device object with a strict schema that has no `authority_scan` field. One env var on one Mac would get all of that Mac's heartbeats rejected, and the Mac would look dead.

## Fix
The marker is sent only once the server has answered the scan flag at least once (`AuthorityScanKnown` from the policy cache, passed through the authority holder). Servers that never answered it never see the field. The report itself was already gated on the server's flag, so a CLI release from `main` now works against current production unchanged.

## Measurement
`TestAuthoritySwitchRecovery` gains a final step: with the flag unknown and the env var set, the heartbeat carries neither report nor marker. `TestAuthorityFlushCadenceAndLastSent` models a server that knows the flag. `go test ./... -race` passes.


hasandemirkiran:
* **#512** <a href="https://app.graphite.com/github/pr/kontext-security/kontext/512?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/kontext-security/kontext/512?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#511** <a href="https://app.graphite.com/github/pr/kontext-security/kontext/511?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#513](https://github.com/kontext-security/kontext/pull/513) <a href="https://app.graphite.com/github/pr/kontext-security/kontext/513?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

